PR #5, opened 2022-02-24
fix: add group to each section

kenji-miyake:
In order to prevent unnecessary propagation of LaunchConfigurations, I added the `group` tag to each section.


kenji-miyake:
I've tested this in my local environment.

yukkysaito (approved):

PR #6, opened 2022-02-24
fix: fix launch args

kenji-miyake:
Adjusted the interface to https://github.com/autowarefoundation/autoware.universe/pull/443.

### DIFF
--- a/autoware_launch/launch/logging_simulator.launch.xml
+++ b/autoware_launch/launch/logging_simulator.launch.xml
@@ -42,11 +42,14 @@
 
   <!-- Vehicle -->
   <group>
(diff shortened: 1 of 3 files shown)


autoware_launch/launch/logging_simulator.launch.xml
```diff
@@ -42,11 +42,14 @@
 
   <!-- Vehicle -->
   <group>
+    <let name="launch_vehicle_interface" value="false" if="$(var vehicle_simulation)"/>
+    <let name="launch_vehicle_interface" value="true" unless="$(var vehicle_simulation)"/>
```
yukkysaito, line 46: IMO, you can make launch_vehicle_interface directly instead of vehicle_simulation

yukkysaito (approved):
LGTM

PR #290, opened 2023-04-11
test(static_centerline_optimizer): add launch test

takayuki5168:
## Description

We cannot notice the bug when launching [static_centerline_optimizer](https://autowarefoundation.github.io/autoware.universe/main/planning/static_centerline_optimizer/) from autoware_launch.
e.g.) The following bug was put into autoware.universe without anyone knowing it resulting in failure to launch static_centerline_optimizer from autoware_launch.
https://github.com/autowarefoundation/autoware.universe/pull/3360

To keep launching static_centerline_optimizer without failure, I added a test.

## Tests performed

launch test passed.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

- [ ] The PR follows the [pull request guidelines].

- [ ] There are no open discussions or they are tracked via tickets.

### DIFF
--- a/autoware_launch/CMakeLists.txt
+++ b/autoware_launch/CMakeLists.txt
@@ -7,6 +7,16 @@ ament_auto_find_build_dependencies()
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
+
+  # static centerline optimizer
+  add_launch_test(
+    test/static_centerline_optimizer/test_static_centerline_optimizer.test.py
+    TIMEOUT "30"
+  )
+  install(DIRECTORY
+    test/static_centerline_optimizer/data/
+    DESTINATION share/${PROJECT_NAME}/test/static_centerline_optimizer/data/
+  )
 endif()
 
 ament_auto_package(
--- a/autoware_launch/package.xml
+++ b/autoware_launch/package.xml
@@ -27,6 +27,7 @@
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>autoware_lint_common</test_depend>
+  <test_depend>ros_testing</test_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
--- a/autoware_launch/test/static_centerline_optimizer/data/lanelet2_map.osm
+++ b/autoware_launch/test/static_centerline_optimizer/data/lanelet2_map.osm
@@ -1,0 +1,146 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<osm generator="VMB">
+  <MetaInfo format_version="1" map_version="2"/>
+  <node id="10" lat="35.62554985786315" lon="139.78246505089274">
+    <tag k="mgrs_code" v="54SUE897431"/>
+    <tag k="local_x" v="89747.3792"/>
+    <tag k="local_y" v="43100.1597"/>
+    <tag k="ele" v="12.149"/>
+  </node>
+  <node id="11" lat="35.625753282567665" lon="139.78285100261152">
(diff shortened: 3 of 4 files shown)


takayuki5168:
@kenji-miyake This is the first test for autoware_launch.

Currently, since many people in TIER IV or AWF run planning_simulator or logging_simulator on a daily basis, we can notice the bug put into autoware.unvierse/autoware_launch when **launching autoware.launch.xml** as soon as possible.
Therefore, no test for launching autoware.launch.xml works.

However, static_centerline_optimizer.launch.xml is used for a few specific people, and if the bug was put into autoware.universe/autoware_launch regarding static_centerline_optimizer, we cannot notice it as soon as possible.
Therefore, I added a launch test for launching static_cenerline_optimizer.

autoware_launch/CMakeLists.txt
```diff
@@ -7,6 +7,16 @@ ament_auto_find_build_dependencies()
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
+
+  # static centerline optimizer
+  add_launch_test(
+    test/static_centerline_optimizer/test_static_centerline_optimizer.test.py
```
kenji-miyake, line 13: [Question] Why is it impossible to test this in the `static_centerline_optimizer` package?

takayuki5168: Sorry to be late.

Do you mean this one?
In the static_centerline_optimizer, the node itself is tested.
https://github.com/autowarefoundation/autoware.universe/blob/main/planning/static_centerline_optimizer/test/test_static_centerline_optimizer.test.py

FYI: In autoware_launch, we should test not the node but launch.xml to launch the node which designates yaml file. That's why I make this PR draft.

kenji-miyake: @takayuki5168 Thank you for the information, but I'm sorry it's still unclear to me.
- What is `launch.xml to launch the node which designates yaml file` in this case? Could you make it a bit more specific?
- What's the difference between https://github.com/autowarefoundation/autoware.universe/blob/main/planning/static_centerline_optimizer/test/test_static_centerline_optimizer.test.py and [autoware_launch/test/static_centerline_optimizer/test_static_centerline_optimizer.test.py](https://github.com/autowarefoundation/autoware_launch/pull/290/files#diff-17f171f631da095225016d0d148f894927f705c83b8cfe444f551c9f708792ac)? They look to be almost the same.

takayuki5168: @kenji-miyake 

> What is launch.xml to launch the node which designates yaml file in this case? 

This one.
https://github.com/autowarefoundation/autoware.universe/blob/main/planning/static_centerline_optimizer/launch/static_centerline_optimizer.launch.xml

> What's the difference

Currently, the difference is only whether launch config or node config is used as a yaml file.
Anyway, the current PR is under edit, so it would be better to discuss again after I fix. Sorry about that.

kenji-miyake: > This one.
https://github.com/autowarefoundation/autoware.universe/blob/main/planning/static_centerline_optimizer/launch/static_centerline_optimizer.launch.xml

Are you trying to test this launch file? If so, where is it called in this pull request?

> Currently, the difference is only whether launch config or node config is used as a yaml file from the test.
Anyway, the current PR is under edit, so it would be better to discuss again after I fix. Sorry about that.

Okay. I think it's good to ask for a review after it's somewhat ready. If it's still a very draft, we should discuss it on an issue.

takayuki5168:
Let me close the PR since with a test in autoware_launch, the problem will not be fixed.

Issue #728, opened 2023-12-08
Duplicate Declaration of Controller Parameter Paths in tier4_control_component.launch.xml


xuyuan-han:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

In the `tier4_control_component.launch.xml` file, the parameters `lat_controller_param_path` and `lon_controller_param_path` are declared twice when including `control.launch.py`. This causes issues with customizing the controller parameter file paths as per user requirements.

### Expected behavior

The user should be able to specify custom paths for the controller parameters, which should be respected by the launch file. The parameters `lat_controller_param_path` and `lon_controller_param_path` should only be declared once, allowing for user-defined values to be effectively passed to `control.launch.py`.

### Actual behavior

Currently, due to the duplicate declaration, the first definition of `lat_controller_param_path` and `lon_controller_param_path` is _overridden_ by the second declaration within the `control.launch.py` include (I am not sure if it is overridden or may lead to undefined behavior). This prevents users from specifying their custom controller parameter paths.

### Steps to reproduce

1. Open the `tier4_control_component.launch.xml` file.
2. Observe the parameter declarations for `lat_controller_param_path` and `lon_controller_param_path`.
3. Notice that these parameters are declared twice – once with a path dependent on `lateral_controller_mode`/`longitudinal_controller_mode` and again later with values from `$(var lat_controller_param_path)`/`$(var lon_controller_param_path)`.
```
<?xml version="1.0"?>
<launch>
  ...

  <include file="$(find-pkg-share tier4_control_launch)/launch/control.launch.py">
    ...

    <arg
      name="lat_controller_param_path"
      value="$(find-pkg-share autoware_launch)/config/control/trajectory_follower/$(var latlon_controller_param_path_dir)/lateral/$(var lateral_controller_mode).param.yaml"
    /> <!-- 1st declaration -->
    <arg
      name="lon_controller_param_path"
      value="$(find-pkg-share autoware_launch)/config/control/trajectory_follower/$(var latlon_controller_param_path_dir)/longitudinal/$(var longitudinal_controller_mode).param.yaml"
    /> <!-- 1st declaration -->
    ...

    <arg name="lat_controller_param_path" value="$(var lat_controller_param_path)"/> <!-- 2nd declaration -->
[...]

PR #940, opened 2024-03-31
fix(tier4_control_component_launch): fix duplicate declaration of controller parameter paths

xuyuan-han:
## Description

see https://github.com/autowarefoundation/autoware_launch/issues/728

## Tests performed

In `autoware.launch.xml`:

```launch
  <arg name="lat_controller_param_path" default="<path_to_arbitrary_lateral_controller_param_file>" description="lateral controller configuration file location"/>
  <arg name="lon_controller_param_path" default="<path_to_arbitrary_longitudinal_controller_param_file>" description="longitudinal controller configuration file location"/>
  <!-- Control -->
  <group if="$(var launch_control)">
    <include file="$(find-pkg-share autoware_launch)/launch/components/tier4_control_component.launch.xml"/>
      <arg name="lat_controller_param_path" value="$(var lat_controller_param_path)"/>
      <arg name="lon_controller_param_path" value="$(var lon_controller_param_path)"/>
  </group>
```

## Effects on system behavior

After fixing, arbitrary lateral and longitudinal controller parameter file paths can be specified in the group `Control` in `autoware.launch.xml` and can be passed to `tier4_control_component.launch.xml`

- [X] I've confirmed the [contribution guidelines].
- [X] The PR follows the [pull request guidelines].

- [ ] The PR follows the [pull request guidelines].

- [X] There are no open discussions or they are tracked via tickets.

### DIFF
--- a/autoware_launch/launch/components/tier4_control_component.launch.xml
+++ b/autoware_launch/launch/components/tier4_control_component.launch.xml
@@ -32,8 +32,6 @@
     <arg name="nearest_search_param_path" value="$(find-pkg-share autoware_launch)/config/control/common/nearest_search.param.yaml"/>
 
     <!-- package param path -->
-    <arg name="lat_controller_param_path" value="$(var lat_controller_param_path)"/>
-    <arg name="lon_controller_param_path" value="$(var lon_controller_param_path)"/>
     <arg name="vehicle_cmd_gate_param_path" value="$(find-pkg-share autoware_launch)/config/control/vehicle_cmd_gate/vehicle_cmd_gate.param.yaml"/>
     <arg name="lane_departure_checker_param_path" value="$(find-pkg-share autoware_launch)/config/control/lane_departure_checker/lane_departure_checker.param.yaml"/>
     <arg name="control_validator_param_path" value="$(find-pkg-share autoware_launch)/config/control/control_validator/control_validator.param.yaml"/>


autoware_launch/launch/components/tier4_control_component.launch.xml
```diff
@@ -10,20 +10,21 @@
...
 
+  <arg
+    name="lat_controller_param_path"
+    default="$(find-pkg-share autoware_launch)/config/control/trajectory_follower/$(var latlon_controller_param_path_dir)/lateral/$(var lateral_controller_mode).param.yaml"
+  />
+  <arg
+    name="lon_controller_param_path"
+    default="$(find-pkg-share autoware_launch)/config/control/trajectory_follower/$(var latlon_controller_param_path_dir)/longitudinal/$(var longitudinal_controller_mode).param.yaml"
+  />
+
```
takayuki5168, line 21: How about removing the following instead?
```xml
  <arg name="lat_controller_param_path" value="$(var lat_controller_param_path)"/>
   <arg name="lon_controller_param_path" value="$(var lon_controller_param_path)"/>
```

takayuki5168: Please assign the PR to me once you fixed it.

xuyuan-han (commented):
The changes made in this pull request are intended to fix the issue. By the way, it appears that I don't have the necessary permissions to assign the pull request.

autoware_launch/launch/components/tier4_control_component.launch.xml
xuyuan-han: The changes made in this pull request are intended to fix the issue. By the way, it appears that I don't have the necessary permissions to assign the pull request.

takayuki5168: Yes, I know.
I proposed a better way to fix it. Do you mean my proposal will not fix the issue?

xuyuan-han: So, exposing these arguments in the manner I did is not permitted?

xuyuan-han: If you simply remove these lines, I don't think the arguments can be exposed to the other launch files.
```
   <arg name="lat_controller_param_path" value="$(var lat_controller_param_path)"/>
   <arg name="lon_controller_param_path" value="$(var lon_controller_param_path)"/>
```

takayuki5168: And what you want to do is use `lat_controller_param_path` directly instead of using `lat_controller_param_path_dir`, right?
Then, the PR title `duplicate declaration` is a bit confusing. More precisely, giving  the lat/lon param path from outside makes sense.

xuyuan-han: Absolutely right. I changed the title. 
The `duplicate declaration` problem does indeed exist. What I propose in this PR is adding a little feature by fixing the bug.
Kill two birds with one stone.

takayuki5168: I got it.

Let me check again.
The ideal design is exposing only the lat_controller_param_path_dir. Why does this not work in your case?

xuyuan-han: In our project, we have our own launch package as `ROS overlay` above the `Autoware` workspace. Our launch package contains only `our_launch/launch/planning_simulator.launch.xml` and `our_launch/launch/autoware.launch.xml`. These two launch files call the low-level ROS underlay launch files in `autoware_launch/launch/components
/components` such as `tier4_control_component.launch.xml`. The idea behind this is to minimize file duplication and merge workload, and to exploit the ROS underlay-overlay feature.

So if the whole `lat/lon_param_path` can be exposed, we can keep our own lat/lon param files in `our_launch/config` and let it be passed to the component launch file.

takayuki5168: Then, how about moving your project's lat/long parameter files in the same way as autoware_launch.
This concealing parameter's path is also applied to other components like tier4_planning_component.launch.xml, and if the specific project wants to change this method, we should create a github discussion to discuss on a larger scale.

Or you can copy tier4_control_component.launch.xml for your project and modify it.

xuyuan-han: OK. Currently we copy and modify tier4_control_component.launch.xml for our project.
So I will follow your solution to remove the following instead.
```
   <arg name="lat_controller_param_path" value="$(var lat_controller_param_path)"/>
   <arg name="lon_controller_param_path" value="$(var lon_controller_param_path)"/>
```

takayuki5168 (approved):
Thanks for the fix!

autoware_launch/launch/components/tier4_control_component.launch.xml
xuyuan-han: It is done. Could you please review the code again and approve the PR?

xuyuan-han:
Why does the task build-and-test-differential / clang-tidy-differential (pull_request) fail?

takayuki5168:
@xuyuan-han The required build passed. Now you can merge.

xuyuan-han:
I don't have write access permission to merge the code. Can you help with this? @takayuki5168

takayuki5168:
Okay, I did.

PR #1061, opened 2024-07-09
chore: renamed the centerpoint config files

knzo25:
reference: https://github.com/autowarefoundation/autoware.universe/pull/7421

In the beforementioned PR, the config files for centerpoint had to be changed to comply with some schema rules
(For a particular schema file, all yaml files matching the base name were compared, which caused issues since `centerpoint` and `centerpoint_tiny` share the same base. Changing them to `centerpoint_base` and `centerpoint_tiny_base` resolved the problem).

This PR should be merged with the previous one simultaneously to avoid issues


miursh:
[Q] Do we need to change the names of the parameter files managed by TIER IV's ML packages as well?

knzo25:
@miursh 

If you refer to the `*_ml_package.param.yaml` files, they are unaffected by this change:

```
kenzolobos@SOME_PC_NAME:~/autoware_data/lidar_centerpoint$ ll
total 50456
drwxr-xr-x  2 kenzolobos kenzolobos     4096  4月 30 14:22 ./
drwxrwxr-x 18 kenzolobos kenzolobos     4096  4月 30 15:06 ../
-rw-r--r--  1 root       root            376  4月 26 20:25 centerpoint_ml_package.param.yaml
-rw-r--r--  1 root       root            374  4月 26 20:25 centerpoint_sigma_ml_package.param.yaml
-rw-r--r--  1 root       root            376  4月 26 20:25 centerpoint_tiny_ml_package.param.yaml
```

The files affected were the `$(var model_name).param.yaml` -> `$(var model_name)_base.param.yaml`,
which as far as I understand (correct me if I am wrong), are only in this repository and universe.

https://github.com/tby-udel/autoware.universe/blob/lidar-centerpoint-branch/perception/lidar_centerpoint/launch/lidar_centerpoint.launch.xml#L8-L9

In the companion PR, the only non-chema /documentation change beside file renaming is the following:
image

knzo25:
Update: there was another instance where the previously mentioned change had to be applied. Asked the author of the related PR to reflect that.

PR #1094, opened 2024-07-26
feat(psim)!: preapre settings to launch localization modules on psim

yuki-takagi-66:
## Description
This PR is a part of the following issue:
https://github.com/autowarefoundation/autoware_launch/issues/1100

This PR provides launch settings to launch Localization modules on psim.
By setting `localization_sim_mode` as `pose_twist_estimator`, most of  localization modules without ndt_localizer will be launched.

Note that the frequency of ekf_localizer and vehicle_sim are required to handle carefully. Frequency setting is under analysis and consideration.

universe PR: https://github.com/autowarefoundation/autoware.universe/pull/8212

## Tests performed
psim tests and scenario simulator

https://evaluation.tier4.jp/evaluation/reports/16cc961f-f08b-527a-b2fa-120dd899dd5d?project_id=prd_jt (internal link)

Note that the scenario simulator is not supported by this PR.

## Effects on system behavior

Not applicable.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

- [x] The PR follows the [pull request guidelines].

- [ ] There are no open discussions or they are tracked via tickets.

### DIFF
--- a/autoware_launch/launch/components/tier4_simulator_component.launch.xml
+++ b/autoware_launch/launch/components/tier4_simulator_component.launch.xml
@@ -51,7 +51,10 @@
       name="object_recognition_tracking_multi_object_tracker_node_param_path"
       value="$(find-pkg-share autoware_launch)/config/perception/object_recognition/tracking/multi_object_tracker/multi_object_tracker_node.param.yaml"
     />
+    <arg name="localization_error_monitor_param_path" value="$(find-pkg-share autoware_launch)/config/localization/localization_error_monitor.param.yaml"/>
+    <arg name="ekf_localizer_param_path" value="$(find-pkg-share autoware_launch)/config/localization/ekf_localizer.param.yaml"/>
     <arg name="pose_initializer_param_path" value="$(find-pkg-share autoware_launch)/config/localization/pose_initializer.param.yaml"/>
+    <arg name="twist2accel_param_path" value="$(find-pkg-share autoware_launch)/config/localization/twist2accel.param.yaml"/>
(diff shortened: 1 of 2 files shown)


takayuki5168 (approved):

kosuke55 (approved):

shmpwk (approved):

mitsudome-r:
Could you explain why you want to launch Localization when using Psim?
My understanding is that PSim is used to test Planning/Control when Perception/Localization outputs are ideal.
Are there any reasons why we are launching localization modules instead of changing algorithms inside PSim for calculating odometry and acceleration?

My concern is that if a developer launch psim locally with `scenario_simulator=false`, there would be a risk that a bug inside localization modules (e.g., ekf) would impact the result of planning/control modules, which would ruin the purpose of psim.

yukkysaito (changes requested):
I will block it to prevent it from being merged.

While PSim used the Component Interface, it is concerning that it will now depend on the internal interfaces of Localization. The role of PSim, as I understand it, is to test Planning/Control under the ideal outputs of Localization/Perception. If its role is to be expanded this much, I think it warrants further discussion.

yuki-takagi-66:
@yukkysaito 
Thank you.
As we discussed internally, we need to achieve the following two goals: 
"maintain a simulator environment whose only dependent module is planning/control" and "evaluate planning/control modules, considering the characteristics of localization in a closed loop."
To achieve these two goals, the preceding PR https://github.com/autowarefoundation/autoware_launch/pull/1106 provided a mode-switching feature supporting three or more states.

mitsudome-r (approved):

TakaHoribe (approved):

yukkysaito (approved):

PR #1099, opened 2024-07-30
feat(autonomous_emergency_braking): enable AEB stop in vehicle_cmd_gate and diag_graph_agg

danielsanchezaran:
## Description
Enables AEB module emergency stop

## Tests performed
PSim

Not applicable.

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #1265, opened 2024-12-09
feat(MRM_handler, MRM_emergency_stop_operator): revert mrm_stop parameter, enable mrm_comfortable_stop

yuki-takagi-66:
## Description
This PR contains the following changes.

1. Revert the mrm_emergency_stop parameter, which was changed by https://github.com/autowarefoundation/autoware_launch/pull/1099 half a year ago. These changes are oriented to AEB; however, these parameters are used for all MRM emergency stops. Therefore, I propose to revert these parameters to maintain any behavior other than AEB.
2. Enabling comfortable_stop. MRM comfortable stop was tested enough by tier4 operations. Now I recommend enabling it as public.

## How was this PR tested?
psim test

## Notes for reviewers
None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/system/mrm_emergency_stop_operator/mrm_emergency_stop_operator.param.yaml
+++ b/autoware_launch/config/system/mrm_emergency_stop_operator/mrm_emergency_stop_operator.param.yaml
@@ -1,5 +1,5 @@
 /**:
   ros__parameters:
     update_rate: 30
-    target_acceleration: -3.0
-    target_jerk: -3.0
+    target_acceleration: -2.5
+    target_jerk: -1.5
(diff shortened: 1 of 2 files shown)


isamu-takagi:
> These changes are oriented to AEB; however, these parameters are used for all MRM emergency stops. Therefore, I propose to revert these parameters to maintain any behavior other than AEB.

@yuki-takagi-66 Can I assume that AEB will still operate safely if I change the parameters back?

yuki-takagi-66:
> > These changes are oriented to AEB; however, these parameters are used for all MRM emergency stops. Therefore, I propose to revert these parameters to maintain any behavior other than AEB.
> 
> @yuki-takagi-66 Can I assume that AEB will still operate safely if I change the parameters back?

@isamu-takagi 
Yes, AEB works properly even if we change the parameters back.
Note that, the complete safety is not our (OSS) aims and this is not achieved by the both parameters.

PR #1307, opened 2025-01-22
feat(autoware_detected_object_validation): add height filter in lanelet filtering

YoshiRi:
## Description
Enable height filter in https://github.com/autowarefoundation/autoware.universe/pull/10003.

## How was this PR tested?

Local build and running.

## Notes for reviewers

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

PR #1358, opened 2025-03-12
chore(autoware_launch): increase default speed limit 

liuXinGangChina:
## Description

Currently the max speed that test vehicle can reach has been limited to 30 km/h. It is impossible for developer to test autoware under higher speed.

we increase default speed limit to 72km/h thus developer can test higher speed scenario both in planning simulator or AWSIM, 

## How was this PR tested?

1. planning simulator
2. AWSIM

Screenshot from 2025-03-12 14-12-42

## Notes for reviewers

None.

## Effects on system behavior

None.


liuXinGangChina:
After discussion with mits san，will close this one
Instead we will make a comment to tell developer how to unleash autoware's full power by oncrease max speed

PR #1359, opened 2025-03-12
feat(pid_long): disable overshoot_emergency

yuki-takagi-66:
## Description
The behavior when crossing the stop line is handled by the control_validator as a reference. For this reason, similar functions on the pid_long side are disabled as default.

## How was this PR tested?
psim

## Notes for reviewers

None.

## Effects on system behavior

None.


kyoichi-sugahara (approved):
This PR is a change to turn off a feature in PID that has become redundant, as [PR](https://github.com/autowarefoundation/autoware.universe/pull/10236) added functionality for the control_validator to publish diagnostics when crossing over stop lines.

**⚠️ NOTICE:** Users who don't have a configuration where the system stops based on the control validator's diagnostics need to be cautious.

PR #872, opened 2024-02-13
feat(map): add dynamic lanelet loading parameters

StepTurtle:
## Description

Dynamically load Lanelet2 maps required for the current driving situation, aiming to avoid time consumption and reduce computational costs in the planning pipeline, while also addressing long-term issues associated with Lanelet2 map projection.

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #1373, opened 2025-03-24
feat: add dynamic lanelet2 loading parameters

mitsudome-r:
## Description
This extracts modification of map.launch.xml out of https://github.com/autowarefoundation/autoware_launch/pull/872.

## Notes for reviewers

None.

## Effects on system behavior

None.


maxime-clem (approved):
Tested and the map dynamic loads as expected (can be visualized by changing the `Lanelet2VectorMap` topic in rviz to `/map/dynamic_vector_map_marker`).

PR #1374, opened 2025-03-24
feat(crosswalk_module): add param to consider objects on crosswalk when pedestrian traffic light is red

mehmetdogru:
## Description

This PR created to add `consider_obj_on_crosswalk_on_red_light` parameter to consider and do not ignore objects if they are on the crosswalk when the crosswalk pedestrian traffic light is red.

:bookmark_tabs:  It is left as `consider_obj_on_crosswalk_on_red_light`: `false` not to change the current system behavior. 

[Related universe PR link](https://github.com/autowarefoundation/autoware_universe/pull/10332)

## Notes for reviewers

None.

## Effects on system behavior

None.


yuki-takagi-66 (approved):
Thank you. LGTM.

PR #1376, opened 2025-03-25
fix(roi_pointcloud_fusion): add roi scale factor param

badai-nguyen:
## Description
- Update parameter for roi_pointcloud_fusion which is added by https://github.com/autowarefoundation/autoware_universe/pull/10333

## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #1379, opened 2025-03-27
fix: update code owners for perception and sensing components

technolojin:
## Description

Update code owners for perception and sensing.

- remove Minoda-san
- add Gu-san and Saeki-san for perception
- add Lee and Kenzo-san for sensing

## Notes for reviewers

None.

## Effects on system behavior

None.


knzo25 (approved):
LGTM

PR #1390, opened 2025-04-08
fix: add missing default param for concatenation

badai-nguyen:
## Description

- Add default parameter for concatenation nodes in ground_segmentation pipeline updated by https://github.com/autowarefoundation/autoware_universe/pull/10411

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/perception/obstacle_segmentation/ground_segmentation/ground_segmentation.param.yaml
+++ b/autoware_launch/config/perception/obstacle_segmentation/ground_segmentation/ground_segmentation.param.yaml
@@ -35,3 +35,20 @@
 
         # debug parameters
         publish_processing_time_detail: false
+    # default parameters for concatenation node when addition_lidars or ransac inputs are used
+    concatenation:
+      parameters:
+        debug_mode: false
+        has_static_tf_only: false
+        rosbag_length: 10.0
+        maximum_queue_size: 5
+        timeout_sec: 0.2
+        is_motion_compensated: false
+        publish_synchronized_pointcloud: false
+        keep_input_frame_in_synchronized_pointcloud: true
+        publish_previous_but_late_pointcloud: false
+        synchronized_pointcloud_postfix: pointcloud
+        input_twist_topic_type: twist
+        output_frame: base_link
+        matching_strategy:
+          type: advanced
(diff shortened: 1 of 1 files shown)


autoware_launch/config/perception/obstacle_segmentation/ground_segmentation/ground_segmentation.param.yaml
```diff
@@ -35,3 +35,20 @@
...
+        timeout_sec: 0.2
+        is_motion_compensated: false
+        publish_synchronized_pointcloud: false
+        keep_input_frame_in_synchronized_pointcloud: true
+        publish_previous_but_late_pointcloud: false
+        synchronized_pointcloud_postfix: pointcloud
+        input_twist_topic_type: odom
+        output_frame: base_link
+        matching_strategy:
+          type: advanced
```
vividf, line 54: The parameters `lidar_timestamp_offsets` and `lidar_timestamp_noise_window` are required for advanced mode.
Could you please add them? 😊 Thank you!

For reference, you can check the following file to ensure they meet your requirements:
https://github.com/tier4/aip_launcher/blob/tier4/universe/aip_x2_gen2_launch/config/concatenate_and_time_sync_node.param.yaml

vividf (changes requested):

autoware_launch/config/perception/obstacle_segmentation/ground_segmentation/ground_segmentation.param.yaml
badai-nguyen: @vividf btw, do you have any recommend for `matching_strategy`? FYI: the timestamp of concatenation input topics are already synchronized.

vividf: If all of the pointclouds already have identical timestamps, please set the lidar_timestamp_offsets to 0.0.

Also, since I'm not very familiar with this node, I noticed that the input_topics parameter is missing.
Was this intentional?

badai-nguyen: > Also, since I'm not very familiar with this node, I noticed that the input_topics parameter is missing. Was this intentional?

yes it was intentional, as the input_topics are dynamic and calculated each time inside `ground_segmentation.launch.py` as https://github.com/autowarefoundation/autoware_universe/pull/10411/files#diff-ab1bf3adbaf51b647105740b912f9a0ec4eb445ea94366d3e41a17b5289a7256R158-R161

vividf:
@badai-nguyen 
I don't think switching from the advanced strategy to the naive one is a good idea in this case, since the naive strategy only considers the arrival time of the pointclouds.

PR #1417, opened 2025-04-23
feat(autoware_launch): use adapi diagnostics

isamu-takagi:
## Description

**This PR should be merged after https://github.com/autowarefoundation/autoware_universe/pull/10539.**
Use diagnostics from adapi node instead of the [component_state_diagnostics](https://github.com/autowarefoundation/autoware_universe/blob/main/system/autoware_system_diagnostic_monitor/script/component_state_diagnostics.py) script.

## How was this PR tested?

Check that the vehicle drives in autonomous mode using simple planning simulator.

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/system/diagnostics/localization.yaml
+++ b/autoware_launch/config/system/diagnostics/localization.yaml
@@ -14,8 +14,8 @@ units:
 
   - path: /autoware/localization/state
     type: diag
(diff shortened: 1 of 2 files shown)


ito-san (approved):
LGTM

PR #1418, opened 2025-04-23
chore: enforce valid perception modes

knzo25:
## Description

We accept a list of perception modes.
However, when the user selects a nonexistent one it would display no error.
This PR forces the perception mode to be an element of the list or fail the launch

## How was this PR tested?

Autoware launchs

## Notes for reviewers

None.

## Effects on system behavior

None.


knzo25:
Related PR: https://github.com/autowarefoundation/autoware_universe/pull/10543

PR #1420, opened 2025-04-25
feat(mission_planner): disable check_footrpint_inside_lanes

kosuke55:
## Description

This feature was introduced by 
https://github.com/autowarefoundation/autoware_universe/pull/2088
but thare are false postives like,

```
[openscenario_interpreter_node-3] [component_container_mt-29] [WARN] [1745174339.543985854] [planning.mission_planning.mission_planner]: Goal's footprint exceeds lane!
```

https://github.com/user-attachments/assets/9ddc6f08-a6f7-4106-a2de-f26b07dc4b40

(this map is private so hard to share)

I think it's better to disable it until fixing false positive.

## How was this PR tested?

2025/04/25 https://evaluation.tier4.jp/evaluation/reports/2d98e3e3-79f8-5fac-81f4-2ced51b576c6/?project_id=prd_jt

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/planning/mission_planning/mission_planner/mission_planner.param.yaml
+++ b/autoware_launch/config/planning/mission_planning/mission_planner/mission_planner.param.yaml
@@ -9,5 +9,5 @@
     reroute_time_threshold: 10.0
     minimum_reroute_length: 30.0
     consider_no_drivable_lanes: false # This flag is for considering no_drivable_lanes in planning or not.
-    check_footprint_inside_lanes: true
+    check_footprint_inside_lanes: false
     allow_reroute_in_autonomous_mode: true


kosuke55:
I will check more

autoware_launch/config/planning/mission_planning/mission_planner/mission_planner.param.yaml
```diff
@@ -9,5 +9,5 @@
     reroute_time_threshold: 10.0
     minimum_reroute_length: 30.0
     consider_no_drivable_lanes: false # This flag is for considering no_drivable_lanes in planning or not.
-    check_footprint_inside_lanes: true
+    check_footprint_inside_lanes: false
```
takayuki5168, line 12: @kosuke55 
Are there any other a lot of false positives?
At least, the case in the description seems easy to be fixed to consider the priority of the orientation to the neighbour lanelets.

mehmetdogru:
@kosuke55 cc. @takayuki5168 

Maybe as a suggestion; among all the lanelets which goal pose is in, the one that goal footprint is within the lanelet boundary could be prioritized.

autoware_launch/config/planning/mission_planning/mission_planner/mission_planner.param.yaml
kosuke55: @takayuki5168 
I haven't analyzed it in detail, but I've seen many false positives so far.
If we fix the case in this video, it might also fix those other cases.
However, lanes are often virtual lines like the video case, so I thought it might be okay to stop checking by default.
The fix of vide case seems easy, but I wasn't keen on allocating the necessary resources.

PR #1428, opened 2025-05-02
feat(radar): update radar pipeline

technolojin:
## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi:
I think this PR should test after related PR.

PR #1440, opened 2025-05-13
feat(control_validator): add lateral jerk validation

kyoichi-sugahara:
## Description

This PR adds lateral jerk threshould for steering rate validation

necessary change for the [PR](https://github.com/autowarefoundation/autoware_universe/pull/10619)

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/control/control_validator/control_validator.param.yaml
+++ b/autoware_launch/config/control/control_validator/control_validator.param.yaml
@@ -9,6 +9,7 @@
 
     thresholds:
       max_distance_deviation: 1.0
+      lateral_jerk: 10.0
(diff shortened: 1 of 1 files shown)


autoware_launch/config/control/control_validator/control_validator.param.yaml
```diff
@@ -9,6 +9,7 @@
 
     thresholds:
       max_distance_deviation: 1.0
+      lateral_jerk: 10.0
```
yuki-takagi-66, line 12: This value seem different from universe value.

yuki-takagi-66 (approved):

PR #1441, opened 2025-05-14
refactor(planning_validator): Implement plugin structure for planning validator node

mkquda:
## Description

add plugins parameter files and update relevant launch files related to following universe PR:
https://github.com/autowarefoundation/autoware_universe/pull/10571

## How was this PR tested?

See universe PR

## Notes for reviewers

None.

## Effects on system behavior

None.


maxime-clem (approved):

satoshi-ota (approved):
Could you add preset args so that we can select load plugin in preset file?
https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/config/planning/preset/default_preset.yaml

PR #1442, opened 2025-05-15
feat(autoware_launch): remove trajectory_deviation from diagnostic graph

isamu-takagi:
## Description

Reflect https://github.com/autowarefoundation/autoware_universe/pull/10337.
The diag `lane_departure_checker_node: trajectory_deviation` is no longer used.

## How was this PR tested?

Check that the vehicle drives in autonomous mode using planning simulator.

## Notes for reviewers

None.

## Effects on system behavior

None.


ito-san (approved):
LGTM

PR #1466, opened 2025-05-29
feat: add logging.launch.xml

takayuki5168:
## Description

To show the diagnostics which is connected to MRM on the rviz and terminal, added logging.launch.xml and updated rviz config file.

image
image

## Notes for reviewers

None.

## Effects on system behavior

None.


mitsudome-r:
I would rather would like to keep the default config of RViz simple.
We can turn it off by default and people can turn on the checkbox for the text overlay if they wish to see the datails.

takayuki5168:
@mitsudome-r 
I agree, and changed it.
https://github.com/autowarefoundation/autoware_launch/pull/1466/commits/f61e696b1a3e15ebb7cca535b624242e0cee02bf

isamu-takagi (approved):

PR #1469, opened 2025-05-30
feat(launch/components): launch in a component-wise manner

youtalk:
## Description

Note that this is for the `humble` branch. After it is merged, it is planned to be applied to the `main` branch as well.

In https://github.com/autowarefoundation/autoware/pull/6184#issuecomment-2914711349, I'm trying to launch each component individually using the launch files under the `launch/components`.

However, when launching each component separately, the global parameter behavior (e.g., `use_sim_time` and `vehicle_model`) provided by `autoware_global_parameter_loader` cannot be reproduced.

https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/launch/autoware.launch.xml#L53-L59
https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_global_parameter_loader/launch/global_params.launch.py#L27

~This PR adds a `component_wise_launch` argument to launch `autoware_global_parameter_loader` if it is `true`. The default is `false`, so there is no change to the behavior when launching from the usual `autoware.launch.xml`.~
This PR runs the `autoware_global_parameter_loader` even if launching each component individually.

## How was this PR tested?

https://github.com/autowarefoundation/autoware/pull/6184#issuecomment-2921550455

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/launch/autoware.launch.xml
+++ b/autoware_launch/launch/autoware.launch.xml
@@ -109,6 +109,8 @@
       <arg name="module_preset" value="$(var planning_module_preset)"/>
       <arg name="enable_all_modules_auto_mode" value="$(var enable_all_modules_auto_mode)"/>
       <arg name="is_simulation" value="$(var is_simulation)"/>
+      <arg name="use_sim_time" value="$(var use_sim_time)"/>
+      <arg name="vehicle_model" value="$(var vehicle_model)"/>
     </include>
   </group>
 
@@ -116,12 +118,17 @@
   <group if="$(var launch_control)">
     <include file="$(find-pkg-share autoware_launch)/launch/components/tier4_control_component.launch.xml">
       <arg name="module_preset" value="$(var control_module_preset)"/>
+      <arg name="use_sim_time" value="$(var use_sim_time)"/>
+      <arg name="vehicle_model" value="$(var vehicle_model)"/>
     </include>
   </group>
 
   <!-- API -->
   <group if="$(var launch_api)">
-    <include file="$(find-pkg-share autoware_launch)/launch/components/tier4_autoware_api_component.launch.xml"/>
+    <include file="$(find-pkg-share autoware_launch)/launch/components/tier4_autoware_api_component.launch.xml">
+      <arg name="use_sim_time" value="$(var use_sim_time)"/>
+      <arg name="vehicle_model" value="$(var vehicle_model)"/>
+    </include>
   </group>
 
   <!-- Tools -->


autoware_launch/launch/components/tier4_autoware_api_component.launch.xml
```diff
@@ -3,5 +3,23 @@
...
+  <arg
+    name="component_wise_launch"
+    default="false"
+    description="Launch in a component-wise manner.
+    If true, this launch file will be launched in a component-wise manner.
+    If false, this launch file will be launched in the autoware.launch.xml."
+  />
+
+  <!-- Global parameters -->
+  <group if="$(var component_wise_launch)" scoped="false">
```
mitsudome-r, line 17: Do you know what happens if we launch global_params.launch.py in both component launch file and top level autoware.launch.xml? Does it produce some kind of an error?

If the behavior won't change, maybe we could remove the component_wise_launch argument.

youtalk: https://github.com/autowarefoundation/autoware_launch/pull/1469/commits/51aba6a1453c2320e493858099561a4e9d3067fd

mitsudome-r:
Also, we should probably modify [autoware.launch.xml](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/launch/autoware.launch.xml) to pass use_sim_time and vehicle_model arguments to loaded component launch files.

youtalk:
@mitsudome-r 
> Also, we should probably modify [autoware.launch.xml](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/launch/autoware.launch.xml) to pass use_sim_time and vehicle_model arguments to loaded component launch files.

Fixed. https://github.com/autowarefoundation/autoware_launch/pull/1469/commits/6959dddbb291c0782141d3acf49006d8a71d61eb

> @youtalk Are we only adding this to api, control, and planning component launch for now? I think that would be enough for launching planning simulator to resolve [this issue](https://github.com/autowarefoundation/autoware/pull/6184#issuecomment-2914711349), but I think we eventually need to add it to the other launch files as well (like localization and perception). Do we plan to do it in a separate PR afterwards?

Since only the `vehicle` component lacks a `launch` file in `autoware_launch` and is only available in `autoware_universe/launch`, the scope of modifications and testing increases. For now, I will keep this PR limited to the minimum necessary changes.

mitsudome-r (approved):

PR #1477, opened 2025-06-05
feat(perception): add parameters to use pedestrian traffic signal result estimated in perception pipeline

MasatoSaeki:
## Description

See https://github.com/autowarefoundation/autoware_universe/pull/10763

## How was this PR tested?

See  https://github.com/autowarefoundation/autoware_universe/pull/10763

## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #1481, opened 2025-06-06
feat(autoware_multi_object_tracker): unknown as static object

technolojin:
## Description
1. let user select if estimate motion (velocity) of unknown object or not.
2. adjust object shape (footprint) to be aligned to the input detected object

## How was this PR tested?

* before
try to estimate its shape, extrapolate by time. therefore, the cluster and the tracked object shape are not matched

https://github.com/user-attachments/assets/503cd104-2bba-4acb-a9c8-4b7a57ae4915

* after (with velocity estimation)
it tracks position and velocity, but the shape is originated by the latest measurement. no shape extrapolation by time.
the cluster and the tracked object shape are matched all the time

https://github.com/user-attachments/assets/2c28ddf8-8777-412c-92b3-2ed1fa1304f3

* after (static motion model)
it dose not estimate velocity. 
the cluster and the tracked object shape are matched all the time

https://github.com/user-attachments/assets/3ba42ff0-feb2-4c52-bc7d-f8de1a3b6d18

## Notes for reviewers

https://github.com/autowarefoundation/autoware_universe/pull/10771

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/perception/object_recognition/tracking/multi_object_tracker/multi_object_tracker_node.param.yaml
+++ b/autoware_launch/config/perception/object_recognition/tracking/multi_object_tracker/multi_object_tracker_node.param.yaml
@@ -15,7 +15,7 @@
     ego_frame_id: base_link
     enable_delay_compensation: true
     consider_odometry_uncertainty: false
-    enable_unknown_object_velocity_estimation: true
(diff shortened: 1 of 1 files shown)


ogrex (approved):
LGTM

PR #1492, opened 2025-06-13
fix(autoware_static_obstacle_avoidance): fix classification method of unstable object

yhisaki:
## Description

- https://github.com/autowarefoundation/autoware_universe/pull/10813

## Notes for reviewers

None.

## Effects on system behavior

None.


satoshi-ota (approved):

autoware_launch/config/planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/static_obstacle_avoidance.param.yaml
```diff
@@ -144,6 +144,8 @@
         merging_vehicle:
           th_overhang_distance: 0.0                     # [m]
 
+        unstable_classification_time: 2.0                # [s] If the object is classified as UNKNOWN type for less than unstable_classification_time seconds, it is probably not an unknown type.
```
satoshi-ota, line 147: [nits] This phrasing might be more appropriate.

```suggestion
        unstable_classification_time: 2.0                # [s] If the object is classified as UNKNOWN for less than unstable_classification_time seconds, we cannot be confident that it truly does not belong to any known class..
```

PR #1513, opened 2025-06-24
Feat/pointcloud container

youtalk:
## Description

When attempting to launch `tier4_sensing.launch.xml` and `tier4_perception.launch.xml` separately in a multi-container setup, there's a problem where they cannot run as composite nodes due to the absence of the `pointcloud_container` that was originally present in `autoware.launch.xml`.
This PR addresses the issue by making each launch file create its own `pointcloud_container`, enabling standalone execution. Note that the behavior remains unchanged when launching using the default `autoware.launch.xml` file.

## How was this PR tested?

- https://github.com/youtalk/autoware/pull/211

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/launch/components/tier4_perception_component.launch.xml
+++ b/autoware_launch/launch/components/tier4_perception_component.launch.xml
@@ -1,5 +1,16 @@
 <?xml version="1.0"?>
 <launch>
+  <arg name="launch_pointcloud_container" default="false" description="if true, launch pointcloud container"/>
+  <arg name="pointcloud_container_name" default="pointcloud_container" description="name of pointcloud container"/>
+
+  <!-- Pointcloud container -->
+  <group if="$(var launch_pointcloud_container)">
+    <include file="$(find-pkg-share autoware_launch)/launch/pointcloud_container.launch.py">
+      <arg name="use_multithread" value="true"/>
+      <arg name="container_name" value="$(var pointcloud_container_name)"/>
+    </include>
+  </group>
+
   <arg name="use_empty_dynamic_object_publisher" default="false" description="if true, /perception/object_recognition/objects topic has an empty DynamicObjectArray"/>
   <arg name="use_traffic_light_recognition" default="true" description="use traffic light recognition module, if false, traffic light recognition module will not be launched"/>
 
--- a/autoware_launch/launch/autoware.launch.xml
+++ b/autoware_launch/launch/autoware.launch.xml
@@ -58,12 +58,6 @@
     </include>
   </group>
 
-  <!-- Pointcloud container -->
-  <include file="$(find-pkg-share autoware_launch)/launch/pointcloud_container.launch.py">
-    <arg name="use_multithread" value="true"/>
-    <arg name="container_name" value="$(var pointcloud_container_name)"/>
-  </include>
-
   <!-- Vehicle -->
   <group if="$(var launch_vehicle)">
     <include file="$(find-pkg-share tier4_vehicle_launch)/launch/vehicle.launch.xml">
--- a/autoware_launch/launch/components/tier4_sensing_component.launch.xml
+++ b/autoware_launch/launch/components/tier4_sensing_component.launch.xml
@@ -1,5 +1,22 @@
 <?xml version="1.0"?>
 <launch>
+  <!-- Launch parameters -->
+  <arg name="launch_pointcloud_container" default="true" description="if true, launch pointcloud container"/>
+  <arg name="pointcloud_container_name" default="pointcloud_container" description="name of pointcloud container"/>
+  <arg name="launch_sensing_driver" default="true"/>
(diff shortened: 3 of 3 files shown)


youtalk:
@yukkysaito The default `launch_pointcloud_container` value is false in `tier4_perception_component.launch.xml` while the default `launch_pointcloud_container` value is true in `tier4_sensing_component.launch.xml`. 
This argument has been added specifically for users who want to run only the perception component. When launching in a multi-container setup, both the sensing and perception components will be launched within a single container to enable the IPC.

yukkysaito:
@youtalk Thank you.
If launch_pointcloud_container is set to true in both tier4_perception_component.launch.xml and tier4_sensing_component.launch.xml, wouldn't pointcloud_container.launch.py be called twice and two containers be launched?

I'll also verify the behavior on my side.

mitsudome-r:
> If launch_pointcloud_container is set to true in both tier4_perception_component.launch.xml and tier4_sensing_component.launch.xml, wouldn't pointcloud_container.launch.py be called twice and two containers be launched?

If they are both set to true, then it launches two containers.

Basically, the PR allows the users to choose if they want to launch a single pointcloud containers for whole Autoware system or launch pointcloud containers independently for each components.

By default, it launches single containers as we do now.

If people would like to launch sensing.launch.xml and perception.launch.xml separately (e.g., in different docker containers), they can launch pointcloud containers for each launch files by setting both true. In that case, I think the users should specify `pointcloud_container_name` argument to make sure that two containers have different node names though.

mitsudome-r:
> If we want to launch a single pointcloud container for the entire Autoware system, what should we do? It seems that the corresponding argument has been removed from autoware.launch.xml.

You can just launch autoware.launch.xml with default arguments just like we do today. By default, `launch_pointcloud_container` for perception.launch.xml is false and sensing.launch.xml is true so poincloud container is only launched within sensing.launch.xml. All the component nodes in both sensing and perception will be loaded to that pointcloud container as long as you pass the same pointcloud_container_name to both launch files.

youtalk:
This PR enables three new execution methods below. Originally, only method 1 was supported.

1. Running the full `autoware.launch.xml` file (default, same as before)
2. Running individual component launch files with single-container execution (performance equivalent to 1)
3. Running individual component launch files with multi-container execution (zero-copy communication not possible)

@mitsudome-r @yukkysaito Please review again.

yukkysaito:
@youtalk 

Please respond to the following question:

> If launch_pointcloud_container is set to true in both tier4_perception_component.launch.xml and tier4_sensing_component.launch.xml, wouldn't pointcloud_container.launch.py be called twice and two containers be launched?


If we launch containers from both sensing.launch.xml and perception.launch.xml, won’t they end up with the same container name?
Does it mean that we need to assign different container names when launching containers from both files?

mitsudome-r (commented):
LGTM, but please make sure to answer @yukkysaito -san's questions.

autoware_launch/launch/components/tier4_perception_component.launch.xml
```diff
@@ -1,8 +1,18 @@
 <?xml version="1.0"?>
 <launch>
+  <arg name="launch_pointcloud_container" default="false" description="if true, launch pointcloud container"/>
+  <arg name="pointcloud_container_name" default="pointcloud_container" description="name of pointcloud container"/>
```
youtalk, line 4:
```suggestion
```

autoware_launch/launch/components/tier4_perception_component.launch.xml
```diff
@@ -1,8 +1,18 @@
...
   <arg name="vehicle_model" default="sample_vehicle"/>
 
+  <!-- Pointcloud container -->
+  <group if="$(var launch_pointcloud_container)">
+    <include file="$(find-pkg-share autoware_launch)/launch/pointcloud_container.launch.py">
+      <arg name="use_multithread" value="true"/>
+      <arg name="container_name" value="$(var pointcloud_container_name)"/>
+    </include>
+  </group>
+
```
youtalk, line 15:
```suggestion
```

yukkysaito:
> If we're not considering launching only with tier4_perception.launch.xml, removing the modifications to tier4_perception.launch.xml would also prevent the problem in case 3 from occurring.

@youtalk 
Thanks for checking.
From a foolproofing perspective, I think it's fine to remove the option to launch from tier4_perception.launch.xml.
However, I believe it would be acceptable as long as there's a comment clearly stating that case 3 is not intended to be supported, so that people are aware.

yukkysaito (approved):
LGTM

mitsudome-r (approved):

technolojin:
@youtalk (@YoshiRi )

Is it intended to be remained?
https://github.com/autowarefoundation/autoware_launch/blob/019bdb8969293a4232879f37ce926f2778869e7f/autoware_launch/launch/autoware.launch.xml#L7

PR #1522, opened 2025-06-27
fix(tier4_simulator_component): fix argument name in tier4_simulator_component.launch.xml

TaikiYamada4:
## Description

The `use_point_cloud_container` argument was added to the `tier4_simulator_component.launch.xml` to realize isolated component-wise launching. https://github.com/autowarefoundation/autoware_launch/pull/1479

However, the `tier4_simulator_launch/launch/simulator.launch.xml` in autoware_universe uses `use_pointcloud_container` that doesn't separate "point" and "cloud".
https://github.com/autowarefoundation/autoware_universe/blob/60075cdf11ddb43a4b6c0bfa9c8f59ffb4e6ba19/launch/tier4_simulator_launch/launch/simulator.launch.xml#L27

This PR fixes the `tier4_simulator_component.launch.xml` to make them the same.

## Notes for reviewers

None.

## Effects on system behavior

None.


youtalk (approved):
Good catch!

PR #1531, opened 2025-07-03
fix: remove radar tracker and set tracked object lanelet filter

technolojin:
## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM for codebase

PR #1543, opened 2025-07-09
feat(planning): migrate for changing planning topic name

yukkysaito:
## Description

**Parent issue:**
- https://github.com/autowarefoundation/autoware/issues/6292

When adopting this new planning framework, we will rename the current planning topic(s). Please refer to the following for details. `/planning/scenario_planning/trajectory` to `/planning/trajectory`

https://github.com/orgs/autowarefoundation/discussions/5033#discussioncomment-13704802

## Notes for reviewers

None.

## Effects on system behavior

None.


mitsudome-r:
@xmfcx I don't think we are changing any existing modules at all in this PR. It is only adding a relay which act as a simple replacement for selector module so that we can make /planning/trajectory as an official final output from the planning module launch file. Then, we can modify control modules to subscribe to /planning/trajectory topic in the next step as mentioned in @yukkysaito -san's [comment](https://github.com/orgs/autowarefoundation/discussions/5033#discussioncomment-13704802).

yukkysaito:
@xmfcx Can you approve this PR?

xmfcx (approved):

PR #1546, opened 2025-07-10
feat(tracking): add lidar_centerpoint_short_range configuration for multi-object tracker

technolojin:
## Description

Add centerpoint short range channel to the multi object tracker configuration.

This option is to be used for short range detection to be enabled in multi-channel mode.

Please see https://github.com/autowarefoundation/autoware_universe/pull/10956

## Notes for reviewers

None.

## Effects on system behavior

None.


ogrex (approved):
Tested and confirmed locally.LGTM

PR #1568, opened 2025-07-23
fix(object_merger): update launch and add param file

badai-nguyen:
## Description

- Update launch and add parameter file for https://github.com/autowarefoundation/autoware_universe/pull/11042

## Notes for reviewers

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

PR #1589, opened 2025-08-05
fix(tracker): update pedestrian max_area_matrix for large pedestrian

technolojin:
## Description
adjust `max_area_matrix` parameters for large pedestrian (pedestrian with baby car, wheelchair, walking with bicycle)

## Notes for reviewers

None.

## Effects on system behavior

None.


ogrex (approved):
Locally tested, and no abnormal behavior occurred.
LGTM

PR #1607, opened 2025-08-14
feat(autoware_launch): change short-circuit of diag setting

isamu-takagi:
## Description

The short-circuit-and type is deprecated, so use and type instead.
Dependencies will be defined using the dependent field.

## How was this PR tested?

Check autonomous driving in planning simulator.

## Notes for reviewers

None.

## Effects on system behavior

None.


ito-san (approved):
LGTM

PR #1610, opened 2025-08-15
fix(image_based_projection_fusion): Set matching_strategy.type to naive

iwatake2222:
## Notes for reviewers

None.

## Effects on system behavior

None.


iwatake2222:
This PR will be canceled, as there are plans to stop using /diagnostics in fusion_node.

PR #1647, opened 2025-09-19
feat(mult object tracker): option to publish merged object in case of multi-channel mode

technolojin:
## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi (commented):
It seems to be Okay. Did you checked psim? It seems planning simulator also has same variable so it may okay(SRY, I did not built on my laptop yet)

autoware_launch/config/perception/object_recognition/tracking/multi_object_tracker/multi_object_tracker_node.param.yaml
```diff
@@ -36,6 +36,7 @@
       [     9.0, 5.0,   9.0, 9.0,     9.0,        4.0,     3.0,        2.0]
 
     # debug parameters
+    publish_merged_objects: false
```
YoshiRi, line 39: This parameter will be overrided in mot node when `is_simulator` is true.
I think at least we need to write some comments for readers.

technolojin:
> It seems to be Okay. Did you checked psim? It seems planning simulator also has same variable so it may okay(SRY, I did not built on my laptop yet)

In planning simulator, it calls the tracker directly
https://github.com/autowarefoundation/autoware_universe/blob/e48e83904cd19b314e1d707c127ebc996e52c93a/launch/tier4_simulator_launch/launch/simulator.launch.xml#L97-L104
therefore, this change not implied to the psim.

autoware_launch/config/perception/object_recognition/tracking/multi_object_tracker/multi_object_tracker_node.param.yaml
technolojin: 1. This will be described in the universe package. https://github.com/autowarefoundation/autoware_universe/pull/11386/commits/fb924beb0751e8fa59d671cefe17ccc068490ae2
2. This is a debug parameter. Most of user does not need to aware.

PR #1662, opened 2025-10-06
feat: apply agnocast for simulator launch, autoware.launch and sensing

Koichi98:
## Description
This is the initial PR for the application of **Agnocast** to various nodes within **Autoware**.

When utilizing Agnocast, [environment variables such as `LD_PRELOAD` and `AGNOCAST_MEMPOOL_SIZE` need to be configured in each node's launch file.](https://github.com/autowarefoundation/autoware_universe/tree/main/common/autoware_agnocast_wrapper)

To facilitate this, the following two values will be passed from the most upstream launch file to each node's launch file:

* **`use_agnocast`**: A boolean value indicating whether `ENABLE_AGNOCAST` is set to `1`.
* **`agnocast_heaphook_path`**: The path to the `agnocast_heaphook` shared library, which is added to `LD_PRELOAD` when `ENABLE_AGNOCAST` is set to `1`.

We need the values from the very top-level launch file to propagate downstream, even during simulator execution. Therefore, when running `planning_simulator` or `logging_simulator`, these values will be used by `autoware_launch` and passed further down to downstream launch files. When `autoware_launch` is executed directly, it will set `use_agnocast` and `agnocast_heaphook_path` itself and then pass them downstream.

## How was this PR tested?
Verified the application by successfully running logging simulator and planning simulator after applying the changes to the /perception/obstacle_segmentation/pointcloud topic.

## Notes for reviewers
The documentation regarding the use of Agnocast in Autoware is available at the link below.
https://github.com/autowarefoundation/autoware_universe/blob/main/common/autoware_agnocast_wrapper/README.md

Furthermore, the following subsequent PRs will be updating other launch files not covered in this PR to also propagate `use_agnocast` and `agnocast_heaphook_path`.

- https://github.com/autowarefoundation/autoware_launch/pull/1659
- https://github.com/autowarefoundation/autoware_launch/pull/1660
- https://github.com/autowarefoundation/autoware_launch/pull/1661

Example of how 'use_agnocast' and 'agnocast_heaphook_path' are used in each node's launch file.

- https://github.com/autowarefoundation/autoware_universe/pull/11448

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/launch/autoware.launch.xml
+++ b/autoware_launch/launch/autoware.launch.xml
@@ -50,8 +50,8 @@
   <arg name="enable_all_modules_auto_mode" default="false" description="enable all module's auto mode"/>
   <arg name="is_simulation" default="false" description="Autoware's behavior will change depending on whether this is a simulation or not."/>
   <!-- Agnocast -->
-  <let name="use_agnocast" value="true" if="$(eval &quot;'$(env ENABLE_AGNOCAST '')' == '1'&quot;)"/>
-  <let name="use_agnocast" value="false" unless="$(eval &quot;'$(env ENABLE_AGNOCAST '')' == '1'&quot;)"/>
+  <let name="use_agnocast" value="true" if="$(eval &quot;'$(env ENABLE_AGNOCAST 0)' == '1'&quot;)"/>
+  <let name="use_agnocast" value="false" unless="$(eval &quot;'$(env ENABLE_AGNOCAST 0)' == '1'&quot;)"/>
   <include file="$(find-pkg-share autoware_launch)/config/agnocast/agnocast.yaml"/>
 
   <!-- Global parameters -->
--- a/autoware_launch/launch/planning_simulator.launch.xml
+++ b/autoware_launch/launch/planning_simulator.launch.xml
@@ -48,8 +48,8 @@
   />
 
   <!-- Agnocast -->
-  <let name="use_agnocast" value="true" if="$(eval &quot;'$(env ENABLE_AGNOCAST '')' == '1'&quot;)"/>
-  <let name="use_agnocast" value="false" unless="$(eval &quot;'$(env ENABLE_AGNOCAST '')' == '1'&quot;)"/>
+  <let name="use_agnocast" value="true" if="$(eval &quot;'$(env ENABLE_AGNOCAST 0)' == '1'&quot;)"/>
+  <let name="use_agnocast" value="false" unless="$(eval &quot;'$(env ENABLE_AGNOCAST 0)' == '1'&quot;)"/>
   <include file="$(find-pkg-share autoware_launch)/config/agnocast/agnocast.yaml"/>
 
   <group scoped="false">


yukkysaito:
Thank you.
If we set agnocast to be enabled by default, could you please provide the evaluation results and an explanation of how much effect it actually has?

autoware_launch/launch/autoware.launch.xml
```diff
@@ -50,6 +50,18 @@
   <arg name="enable_all_modules_auto_mode" default="false" description="enable all module's auto mode"/>
   <arg name="is_simulation" default="false" description="Autoware's behavior will change depending on whether this is a simulation or not."/>
```
yukkysaito, line 52:
```suggestion

```

mitsudome-r:
https://github.com/orgs/autowarefoundation/discussions/5835#discussioncomment-12413398
I think there were some points that we want to solve before making it a default. Were these points resolved already?

mitsudome-r:
Even if we are integrating agnocast to these launch files, I would like to make the structure much simpler.

Koichi98:
> I think there were some points that we want to solve before making it a default. Were these points resolved already?
@mitsudome-r 
We won't be making Agnocast the default. The changes will actually only be applied and Agnocast utilized when you set 'ENABLE_AGNOCAST=1'. Does this resolve your concerns?

Koichi98:
> Even if we are integrating agnocast to these launch files, I would like to make the structure much simpler.
@mitsudome-r 
I agree. I've deleted an unnecessary variable so that it looks much simpler. 
https://github.com/autowarefoundation/autoware_launch/pull/1662/commits/07fcc56ec5b70242943f54248ab9c6525ba57968
https://github.com/autowarefoundation/autoware_launch/pull/1662/commits/5e275a0d42e82d201203c764f4126c31be036dde
I would appreciate it if you could take a look at the latest version.

autoware_launch/launch/autoware.launch.xml
```diff
@@ -49,6 +49,11 @@
   <!-- Auto mode setting-->
   <arg name="enable_all_modules_auto_mode" default="false" description="enable all module's auto mode"/>
   <arg name="is_simulation" default="false" description="Autoware's behavior will change depending on whether this is a simulation or not."/>
+  <!-- Agnocast -->
+  <let name="use_agnocast_default" value="true" if="$(eval &quot;'$(env ENABLE_AGNOCAST '')' == '1'&quot;)"/>
+  <let name="use_agnocast_default" value="false" unless="$(eval &quot;'$(env ENABLE_AGNOCAST '')' == '1'&quot;)"/>
+  <arg name="use_agnocast" default="$(var use_agnocast_default)" description="enable Agnocast (overrides ENABLE_AGNOCAST env var)"/>
+  <arg name="agnocast_heaphook_path" default="$/opt/ros/humble/lib/libagnocast_heaphook.so" description="path to agnocast heaphook library"/>
```
mitsudome-r, line 56: Is '$' needed, or is this a typo? 
```suggestion
  <arg name="agnocast_heaphook_path" default="/opt/ros/humble/lib/libagnocast_heaphook.so" description="path to agnocast heaphook library"/>
```

Koichi98: typo. I haven't been able to test it since I applied the fix. I'll comment again once I've confirmed it works.
Thank you.

autoware_launch/launch/autoware.launch.xml
```diff
@@ -49,6 +49,11 @@
   <!-- Auto mode setting-->
   <arg name="enable_all_modules_auto_mode" default="false" description="enable all module's auto mode"/>
   <arg name="is_simulation" default="false" description="Autoware's behavior will change depending on whether this is a simulation or not."/>
+  <!-- Agnocast -->
+  <let name="use_agnocast_default" value="true" if="$(eval &quot;'$(env ENABLE_AGNOCAST '')' == '1'&quot;)"/>
+  <let name="use_agnocast_default" value="false" unless="$(eval &quot;'$(env ENABLE_AGNOCAST '')' == '1'&quot;)"/>
+  <arg name="use_agnocast" default="$(var use_agnocast_default)" description="enable Agnocast (overrides ENABLE_AGNOCAST env var)"/>
+  <arg name="agnocast_heaphook_path" default="/opt/ros/humble/lib/libagnocast_heaphook.so" description="path to agnocast heaphook library"/>
```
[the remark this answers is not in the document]
yukkysaito: Also, would it be possible to avoid exposing agnocast-specific parameters in autoware.launch.xml and instead describe them in a YAML file that gets loaded?

PR #1690, opened 2025-11-05
fix(ekf_localizer): change for queue pop on ekf_localizer

Motsu-san:
## Description
This is the parameter change for https://github.com/autowarefoundation/autoware_core/pull/679

## Notes for reviewers

None.

## Effects on system behavior

None.


TaikiYamada4 (approved):
LGTM!

PR #1693, opened 2025-11-07
feat(mission_planner): parameter changes for manual lane change handler

PanConChicharron:
## Description
Parmeter changes for the UI visualisation of preferred-lanes.

Part of https://github.com/autowarefoundation/autoware_universe/pull/11169

**Parent Issue:**

- [Link](https://tier4.atlassian.net/browse/RT1-10751)

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/planning/mission_planning/mission_planner/mission_planner.param.yaml
+++ b/autoware_launch/config/planning/mission_planning/mission_planner/mission_planner.param.yaml
@@ -13,4 +13,4 @@
     consider_no_drivable_lanes: false # This flag is for considering no_drivable_lanes in planning or not.
     check_footprint_inside_lanes: true
     allow_reroute_in_autonomous_mode: true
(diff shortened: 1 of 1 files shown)


autoware_launch/config/planning/mission_planning/mission_planner/mission_planner.param.yaml
```diff
@@ -13,3 +13,4 @@
     consider_no_drivable_lanes: false # This flag is for considering no_drivable_lanes in planning or not.
     check_footprint_inside_lanes: true
     allow_reroute_in_autonomous_mode: true
+    goal_lanelet_transparency: 0.05
```
zulfaqar-azmi-t4: It might be good to add 1 liner comment which describe what it is and the effect if we increase or decrease the value.

Issue #1698, opened 2025-11-11
Release sample_sensor_description and sample_vehicle_description package from ROS Buildfarm


mitsudome-r:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I've agreed with the maintainers that I can plan this task.

### Description

Release sample_sensor_description and sample_vehicle_description package from ROS Buildfarm.

### Purpose

Currently, we have instructions for running Autoware Core with AWSIM, but the users would have to build sample_sensor_description and sample_vehicle_description from source. Having the released as ROS binary package simplifies the steps to run AWSIM with Autoware Core.

https://autowarefoundation.github.io/autoware-documentation/main/tutorials/quick-start/autoware-core-awsim/#launch-autoware-for-debian-package-installation

### Possible approaches

Register the package to ROS Buildfarm.
Since autoware_launch has too many dependencies, we will only focus on releasing the description packages only.

### Definition of done

- [x]  update missing dependencies in package.xml https://github.com/autowarefoundation/autoware_launch/pull/1699
- [ ]  release [sensor_component_description](https://github.com/tier4/sensor_component_description) description package from ROS buildfarm
  - [ ]  move the repo to autowarefoundation
  - [ ]  remove velodyne_description since it is already released from ROS Buildfarm
  - [ ]  register the package to rosdistro
- [ ]  release sample_sensor_description and sample_vehicle_description packages from ROS Buildfarm

PR #1699, opened 2025-11-11
fix: add missing exec_depends for sensor description

mitsudome-r:
## Description
This is part of https://github.com/autowarefoundation/autoware_launch/issues/1698
This PR adds missing dependencies to sensor description packages.

For example, sample_sensor_description has sensor_kit.xacro which depends on vls_descrption, camera_description, and imu_description, but they are not described in package.xml.
https://github.com/autowarefoundation/autoware_launch/blob/8773c132ae3327ed00fabe33e2c9e98e3c05e4bf/sensor_kit/sample_sensor_kit_launch/sample_sensor_kit_description/urdf/sensor_kit.xacro#L4-L7

Also, original package.xml had velodyne_description as `depend`, but it is actually `exec_depend`.
Similar modifications are made to other sensor_description packages.

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/sensor_kit/sample_sensor_kit_launch/sample_sensor_kit_description/package.xml
+++ b/sensor_kit/sample_sensor_kit_launch/sample_sensor_kit_description/package.xml
@@ -10,7 +10,7 @@
 
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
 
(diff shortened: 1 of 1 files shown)


yukkysaito (approved):
LGTM

PR #1711, opened 2025-11-20
fix(run_out): reuse the previous stop pose when possible

maxime-clem:
## Description

Add the parameter required for https://github.com/autowarefoundation/autoware_universe/pull/11681

## How was this PR tested?

Psim + Reproducer

## Notes for reviewers

None.

## Effects on system behavior

None.


autoware_launch/config/planning/scenario_planning/lane_driving/motion_planning/motion_velocity_planner/run_out.param.yaml
```diff
@@ -32,6 +32,7 @@
         keep_condition:  # keep the stop decision if we still find any collision with the object within some time+distance range
           time: 5.0  # [s] time along the ego trajectory
           distance: 10.0  # [m] distance along the ego trajectory
+        reuse_margin: 5.0 # [m] the previous stop distance is reused if before the newly calculated one but not further than this margin
```
zulfaqar-azmi-t4:
```suggestion
        reuse_margin: 5.0 # [m] the previous stop pose is reused if it is positioned before the newly calculated stop pose and their arc length difference is smaller than this margin
```

PR #1718, opened 2025-12-11
feat: allow the latest frame from each tlr camera to stay for a little longer

Owen-Liuyuxuan:
## Description

https://github.com/autowarefoundation/autoware_universe/pull/11769

As discussed in the PR above, the parameter needs to be tuned to about 0.12 for the system to be more stable.

## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi (approved):
This value sorely seems to be okay. Following the universe review status.

Owen-Liuyuxuan:
@MasatoSaeki Thank you, instead of creating new parameter, this PR will now tune the existing parameter.

PR #1719, opened 2025-12-11
feat(autoware_launch): specify api launch args explicitly

isamu-takagi:
## Description

Specify api launch args explicitly.

## How was this PR tested?

Check autonomous driving in planning simulation.

## Notes for reviewers

None.

## Effects on system behavior

None.


sasakisasaki (approved):
Change request is applied. Thank you!

PR #1730, opened 2026-01-06
fix(roi_cluster_fusion): separate iou threshold for each class

badai-nguyen:
## Description
- Update parameter for `roi_cluster_fusion` to align with https://github.com/autowarefoundation/autoware_universe/pull/11828
## Notes for reviewers

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

This PR will be merged with https://github.com/autowarefoundation/autoware_universe/pull/11828 in the same time.

PR #1757, opened 2026-02-16
feat(autoware_launch): add dbt camera cross validation

kaancolak:
## Description

Add a separate `roi_detected_object_fusion` step for `detection_by_tracker` (DBT) objects in the detection pipeline. This filters phantom detections, particularly pedestrians that originate from the tracker feedback loop.

### Problem

In the existing pipeline, the output of `detection_by_tracker` was fed directly into the merger/tracker without any camera validation.Once a momentary false-positive tracker was created for a static object misclassified as a pedestrian, `detection_by_tracker` would continuously find the point cloud cluster at that position and re-publish it as a pedestrian detection. Then `detection_by_tracker` would find the point cloud cluster at that position and re-publish it as a pedestrian detection. Since the DBT channel's `can_spawn_new_tracker`([link](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/config/perception/object_recognition/tracking/multi_object_tracker/input_channels.param.yaml#L97)) is `false` but it can still update existing trackers with a default existence probability of 0.75, the tracker's total existence probability stays above the expiry threshold indefinitely, effectively living forever through the feedback mechanism.

https://github.com/user-attachments/assets/6dca0cdd-c6fb-4ed2-b683-8784febf8953

### Solution

A `roi_detected_object_fusion` node is inserted in detection.launch.xml between the `detection_by_tracker` node output and all downstream consumers (both multi-channel tracker and merged-mode merger). The DBT node now outputs to an intermediate topic (detection_by_tracker/unfiltered/objects), which is validated against camera ROIs before being published to the original topic (detection_by_tracker/objects).

A dedicated parameter file (roi_detected_object_fusion_dbt.param.yaml) is used with stricter thresholds for PEDESTRIAN:

- passthrough_lower_bound_probability_threshold: 1.0 → all DBT pedestrians must be confirmed by camera ROI (no passthrough)
- trust_distance: 70.0m validation range

https://github.com/user-attachments/assets/6b7b02b1-7273-40c1-b0f4-5318617887c9

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/perception/object_recognition/detection/image_projection_based_fusion/roi_detected_object_fusion_dbt.param.yaml
+++ b/autoware_launch/config/perception/object_recognition/detection/image_projection_based_fusion/roi_detected_object_fusion_dbt.param.yaml
@@ -1,0 +1,21 @@
+/**:
+  ros__parameters:
+    # UNKNOWN, CAR, TRUCK, BUS, TRAILER, MOTORBIKE, BICYCLE, PEDESTRIAN
+    # For detection_by_tracker objects: existence_probability = 2D IoU (not NN confidence).
+    # Set PEDESTRIAN to 1.0 so ALL dbt pedestrians must pass camera ROI validation (no passthrough).
+    passthrough_lower_bound_probability_thresholds: [0.35, 0.35, 0.35, 0.35, 0.35, 0.35, 0.35, 1.0]
+    trust_distances: [50.0, 100.0, 100.0, 100.0, 100.0, 50.0, 50.0, 70.0]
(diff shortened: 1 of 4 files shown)


kaancolak:
@technolojin -san, Thanks for the reply and informations.

For point 2, I fixed it over topic manipulation: ROI fusion now has a dual guard (switch/detector/tracker_based AND switch/detector/camera_lidar) and the intermediate topic is only wired in camera modes, so in lidar, lidar_radar, and radar modes DBT outputs directly to detection_by_tracker/objects with no ROI fusion node launched.

PR #1772, opened 2026-02-26
feat(tier4_universe_launch): use control command gate by default

isamu-takagi:
## Notes for reviewers

None.

## Effects on system behavior

None.


isamu-takagi:
https://github.com/autowarefoundation/autoware_universe/issues/12497

PR #1797, opened 2026-03-17
ci: add roscope CI for integrity check

paulsohn:
## Description

This is a demonstration of https://github.com/paulsohn/roscope in autoware_launch CI.

It resolves the actual launcher path executed by `ros2 launch autoware_launch autoware.launch.xml sensor_model:=sampe_sensor_kit vehicle_model:=sample_vehicle map_path:=[map_path]` (map path is just a placeholder now), inlining the includes and evaluating args and variables, to obtain a clear blueprint that what the launcher wants to execute and what packages should be built for it.

See https://github.com/orgs/autowarefoundation/discussions/6870 as well: only with roscope locally, I already found 3 bugs and 2 potential defects.

## How was this PR tested?

https://github.com/paulsohn/roscope/blob/devel/.github/workflows/autoware-integration.yml

roscope repository runs CI frequently to watch if there is any defect on the entire build system.
This CI check gives an identical experience of the autoware_launch repository itself.

## Notes for reviewers

Since the launch system expects ML param files to be in the specified directory, ML parameter files are also added to the PR so they can be copied into the expected directory for the CI check.

`ci_depends.repos` are excerpted from `repositories/autoware.repos` and `repositories/autoware-nightly.repos`, so they should be updated if you want a version change of the launch dependencies (e.g. we recently bumped nebula to v0.3.x).

## Effects on system behavior

None. This is just a CI check.


paulsohn:
https://github.com/paulsohn/autoware_launch/pull/13

e.g. param file not found issue can be caught with this CI check.

xmfcx:
**Follow-up suggestion: automate `ci_depends.repos` syncing**

You already flagged the drift hazard in the PR description ("should be updated when nebula bumps"). To avoid future bit-rot, a scheduled bot PR that diffs `ci_depends.repos` against upstream and opens a sync PR would close the loop nicely.

The auth/branching/PR-creation skeleton from `.github/workflows/bump-version-pr.yaml` (App token → `autoware-github-actions/set-git-config` → `gh pr create`) is directly reusable. Rough shape:

```yaml
name: sync-ci-depends
on:
  schedule: [{cron: "0 5 * * 1"}]  # weekly Monday
  workflow_dispatch:

jobs:
  sync:
    runs-on: ubuntu-24.04
    steps:
      - uses: actions/checkout@v4
      - run: |
          curl -fsSLO https://raw.githubusercontent.com/autowarefoundation/autoware/main/repositories/autoware.repos
          curl -fsSLO https://raw.githubusercontent.com/autowarefoundation/autoware/main/repositories/autoware-nightly.repos
      - run: python3 .github/scripts/sync_ci_depends.py
      # reuse the App-token + set-git-config + gh pr create glue
      # from bump-version-pr.yaml here
```

The small `sync_ci_depends.py` script: for each repo key in local `ci_depends.repos`, look it up in `autoware-nightly.repos` first (to preserve "track \`main\`" semantics for the entries that intentionally do so), fall back to `autoware.repos`, update `version` in place. Roughly 40 lines.

Not blocking this PR.

.github/workflows/roscope-check.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  schedule:
+    - cron: 0 6 * * * # daily 6 AM UTC
+  workflow_dispatch:
+
+concurrency:
```
xmfcx, line 14: Cheap defense in depth: the workflow only needs to read repo contents. Adding an explicit `permissions:` block also documents intent and protects against future org-level default changes.

```suggestion
permissions:
  contents: read

concurrency:
```

xmfcx (approved):
Approving. Real CI gap that no stock ROS 2 tool covers, scope is small and surgical, and CI is green on the current head.

One inline suggestion above (`permissions: contents: read`) is optional defense in depth, not a blocker.

Drift hazards (ci_depends.repos pin sync, vendored ML param staleness) are tracked in the follow-up suggestion comment for a scheduled bot PR. Coverage expansion to other launchers (planning_simulator, logging_simulator, e2e_simulator) and supply-chain hardening for the roscope `.deb` are also good follow-up territory.

Thanks for pushing this through.

.github/workflows/roscope-check.yaml
paulsohn: Cool, by no means I am intending to do some spooky things on this repo, but it will clear everyone else's doubt. Let me check if the integrity check works after restricting the permission.

PR #1751, opened 2026-02-12
feat(autoware_sample_designs): implement autoware system designer systems

technolojin:
## Description

Add the sample system implementation by the autoware_system_designer.

DISCUSSION https://github.com/orgs/autowarefoundation/discussions/6760
ISSUE https://github.com/autowarefoundation/autoware/issues/6793

This PR need to be merged after
- https://github.com/autowarefoundation/autoware/pull/6739
- https://github.com/autowarefoundation/autoware/pull/6985

## How was this PR tested?

To try autoware_system_designer, please follow instruction [here](https://github.com/autowarefoundation/autoware/pull/6739)

## Notes for reviewers

- autoware_common_designs: common module configurations
- autoware_sample_deployment: implementation of the sample system.
  * logging-simulation
  * planning-simulation
  * AWSIM

## Effects on system behavior

None.

PR #1810, opened 2026-04-08
feat(autoware system designs): autoware system designs psim add planning control

technolojin:
## Description

add planning and control component designs.

this PR is base on https://github.com/autowarefoundation/autoware_launch/pull/1751

planning

image

control

image

all (runtime)
 
Screenshot from 2026-05-21 09-12-19

## How was this PR tested?

apply planning and control nodes 
https://github.com/autowarefoundation/autoware_universe/pull/12408

need to update autoware system designer https://github.com/autowarefoundation/autoware_system_designer/pull/58

## Notes for reviewers

None.

## Effects on system behavior

None.


maxime-clem (approved):
Tested and it works well 👍 The only difference I could notice is that Autoware stops at traffic lights if no signal is published, whereas with the default launch system it does not.

PR #1826, opened 2026-04-21
feat: polar voxel noise filter

SergioReyesSan:
## Description

Modification to the launcher to run the simple voxel noise filter. This implementation includes the configuration files and modifications done when testing the autoware_pointcloud_preprocessor.

The configuration is done in a way that the filter could be disabled without altering the lidar preprocessor pipeline operation.

Tested only with rosbags.

## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM but IMO we need some build and running test so that this change does not break awf autoware launching

PR #1829, opened 2026-04-27
feat(autoware_launch): remove MrmSummaryoverlayPlugins 

mitsudome-r:
## Description
This removes MRM summary overlay plugins from rviz configs as the plugin was deleted in https://github.com/autowarefoundation/autoware_universe/pull/11376.

## How was this PR tested?

* `ros2 run rviz2 rviz2 -d $(ros2 pkg prefix --share autoware_launch)/rviz/autoware_test_utils.rviz`
* `ros2 run rviz2 rviz2 -d $(ros2 pkg prefix --share autoware_launch)/rviz/perception.rviz`
* `ros2 run rviz2 rviz2 -d $(ros2 pkg prefix --share autoware_launch)/rviz/planning_bev.rviz`

perception.rviz and planning_bev.rviz still fails due to other plugins that are missing regardless of this PR. I would create the follow up PR to fix them. 

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/rviz/autoware_test_utils.rviz
+++ b/autoware_launch/rviz/autoware_test_utils.rviz
@@ -283,22 +283,6 @@ Visualization Manager:
               Width: 550
           Enabled: true
           Name: Vehicle
-        - Class: rviz_plugins/MrmSummaryOverlayDisplay
-          Enabled: false
-          Font Size: 10
-          Left: 512
-          Max Letter Num: 100
-          Name: MRM Summary
-          Text Color: 25; 255; 240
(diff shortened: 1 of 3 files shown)


soblin (approved):

paulsohn:
@mitsudome-r Should we remove https://github.com/mitsudome-r/autoware_launch/blob/ad74204ae6911846087ba6dac3c5a67fb489267e/autoware_launch/rviz/planning_bev.rviz#L2722 (and two more) as well?

image

PR #1831, opened 2026-04-28
ci: register map param files for sync-params workflow + sync-params bugfix

paulsohn:
## Description

Tracking issue: https://github.com/autowarefoundation/autoware/issues/7048

Sync-params workflow: #1812 

Register map param files with their upstream sources so that param files can be synced (via PR) periodically or by a manual trigger, excluding fields with `{OVERRIDE}` comment marker.

image

So sync-params script itself has been patched because of an error. No regression is detected.

## How was this PR tested?

Not available.

## Notes for reviewers

While I briefly identified the corresponding nodes for each param file, I need the reviewer's help to cross-check whether the mapping from the upstream and downstream files (`.github/sync-params.yaml`, under `map` section) are set correctly.

## Effects on system behavior

None expected. Active parameters remain same.

### DIFF
--- a/.github/sync-params.yaml
+++ b/.github/sync-params.yaml
@@ -124,7 +124,26 @@ perception:
       # Perception/traffic_light_recognition
 
 localization: []
-map: []
+map:
+  - repository: autowarefoundation/autoware_core
+    ref: main
+    files:
+      - source: map/autoware_map_loader/config/lanelet2_map_loader.param.yaml
+        variants:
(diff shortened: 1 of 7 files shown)


.github/sync-params.yaml
KYabuuchi, line 1: The configuration and package correspondence looks good. :+1:

YamatoAndo (approved):

PR #1836, opened 2026-04-30
feat(autoware_launch): make launch of simulator_interface optional for e2e_simulator.launch.xml

mitsudome-r:
## Description
`autoware_carla_interface` not only bridges Autoware and CARLA, but also spawns the world and the ego vehicle inside CARLA as part of its launch. There are use cases where a user wants to relaunch Autoware while leaving the existing CARLA environment (world + ego vehicle) untouched. With this flag set to false, the user can launch autoware_carla_interface individually in other terminal to enable re-running Autoware without re-initializing the CARLA scene.

## Related links

 - https://github.com/autowarefoundation/autoware/issues/7079

## How was this PR tested?
 - Launch with defaults and confirm autoware_carla_interface still starts when simulator_type:=carla
 - Launch with launch_simulator_interface:=false and simulator_type:=carla, verify Autoware comes up without re-initializing the CARLA world/vehicle, and that an already-running CARLA scene continues to work

## Notes for reviewers

I also changed the logics for comparing strings within launch

```
if="$(eval '\'$(var simulator_type)\' == \'carla\'')
```
to
```
if="$(equals $(var simulator_type) 'carla')"
```

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/launch/e2e_simulator.launch.xml
+++ b/autoware_launch/launch/e2e_simulator.launch.xml
@@ -45,7 +45,7 @@
   <group scoped="false" if="$(var launch_simulator_interface)">
     <!-- AWSIM does not require a simulator interface since it publishes autoware_msgs directly -->
     <!-- CARLA -->
-    <group scoped="false" if="$(eval '\'$(var simulator_type)\' == \'carla\'')">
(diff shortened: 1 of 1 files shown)


yukkysaito (approved):
LGTM

PR #1855, opened 2026-06-01
fix: update prameter related to 2 new classess for detection pipeline

badai-nguyen:
## Description
- For the issue: https://github.com/autowarefoundation/autoware/issues/7137
- Need merger together with https://github.com/autowarefoundation/autoware_universe/pull/12671
## Notes for reviewers

None.

## Effects on system behavior

None.


autoware_launch/config/perception/object_recognition/detection/object_merger/data_association_matrix.param.yaml
```diff
@@ -8,44 +8,52 @@
...
+      [0,       0,   0,     0,    0,       0,         0,       0,          0,      0,      #UNKNOWN
+       0,       1,   1,     1,    1,       0,         0,       0,          0,      0,      #CAR
+       0,       1,   1,     1,    1,       0,         0,       0,          0,      0,      #TRUCK
+       0,       1,   1,     1,    1,       0,         0,       0,          0,      0,      #BUS
+       0,       1,   1,     1,    1,       0,         0,       0,          0,      0,      #TRAILER
+       0,       0,   0,     0,    0,       1,         1,       1,          1,      1,      #MOTORBIKE
+       0,       0,   0,     0,    0,       1,         1,       1,          1,      1,      #BICYCLE
+       0,       0,   0,     0,    0,       1,         1,       1,          1,      1,      #PEDESTRIAN
+       0,       0,   0,     0,    0,       1,         1,       1,          1,      1,      #ANIMAL
+       0,       0,   0,     0,    0,       1,         1,       1,          1,      1]      #HAZARD
```
technolojin, line 22: note: this is possible since the object merger measures the matrix(vector) size and square root the vector size to know matrix size.

https://github.com/autowarefoundation/autoware_universe/blob/d0e757a87f26457fd63dd49fab8eda3d6a68a87f/perception/autoware_object_merger/src/association/data_association.cpp#L64

technolojin (approved):
LGTM

paulsohn:
@badai-nguyen

image

check-params is failing. I added this check-param to guarantee the same shape of param files between autoware_universe and autoware_launch. Even if param files in autoware_universe are unused directly, they serve a reference schema for the integration layer. (yes we have json schema, but it doesn't specify matrix width/height)

Apologies that the script is a bit unfriendly, but if a param file schema in autoware_universe has changed, you should either run this workflow

image

or locally run `python3 .github/scripts/sync_params.py perception` to pin the updated version.

***

In https://github.com/autowarefoundation/autoware_launch/pull/1858/changes you can see how CI complains the schema mismatch and tries to roll back.

The CI check failure comes from

* `tracked_object_lanelet_filter.param.yaml`
* `overlapped_judge.param.yaml`
* `ml_camera_lidar_object_association_merger.param.yaml`
* `data_association_matrix.param.yaml`
* `obstacle_pointcloud_based_validator.param.yaml`

two additional labels affect these files in autoware_universe, so they should be updated as well as autoware_launch side.

on the other hand
* `object_lanelet_filter.param.yaml`
* `object_position_filter.param.yaml`
* `tracked_object_lanelet_filter.param.yaml`

needs `ANIMAL` and `HAZARD` but this PR does not contain them.

You should:
1. update 5 leftover param files in autoware_universe so that `ANIMAL` and `HAZARD` appears in the matrix.
2. After merging 1, in this repository run `python3 .github/scripts/sync_params.py perception` and see what happens.

***

For `radar_lanelet_filter.param.yaml`, you will see something like this:

image

it is because the upstream (param file in autoware_universe) value is set to `false`. You can mark `ANIMAL : true # {OVERRIDE}` to override the upstream value, AFTER running the script.

technolojin:
@badai-nguyen 
https://github.com/autowarefoundation/autoware_launch/pull/1858

Since autoware_object_merger package's default configs are not updated, this CI tries to roll back.

PR #1856, opened 2026-06-01
feat: update multi-object tracker parameters for improved shape-class association

technolojin:
## Description
update tracker association parameters to follow update https://github.com/autowarefoundation/autoware_universe/pull/12672

check-param will be fixed after the universe PR is done.

## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #1863, opened 2026-06-04
ci: register planning param files for sync-params workflow (1)

paulsohn:
## Description

Tracking issue: https://github.com/autowarefoundation/autoware/issues/7048

Sync-params workflow: #1812 

Register planning param files with their upstream sources so that param files can be synced (via PR) periodically or by a manual trigger, excluding fields with `{OVERRIDE}` comment marker.

image

This PR covers mission_planner, neural_net_planner, parking, and common part of scenario_planning. Remaining part (behavior_planning, motion_planning) will be added in another PR.

## How was this PR tested?

Not available.

## Notes for reviewers

While I briefly identified the corresponding nodes for each param file, I need the reviewer's help to cross-check whether the mapping from the upstream and downstream files (`.github/sync-params.yaml`, under `planning` section) is set correctly.

## Effects on system behavior

None expected.


autoware_launch/config/planning/mission_planning/mission_planner/mission_planner.param.yaml
```diff
@@ -1,16 +1,29 @@
...
+
 /**:
   ros__parameters:
     map_frame: map
     arrival_check_angle_deg: 45.0
-    arrival_check_duration: 1.0
     arrival_check_lateral_distance: 1.0
     arrival_check_longitudinal_undershoot_distance: 1.0
     arrival_check_longitudinal_overshoot_distance: 2.0
+    arrival_check_duration: 1.0
```
shmpwk, line 21: [Nits] Why do we need this change

paulsohn: The sync-param behavior prioritizes the upstream parameter order and styles, which might contain the upstream intent to group related parameters.

shmpwk: Understood

shmpwk (approved):
> While I briefly identified the corresponding nodes for each param file, I need the reviewer's help to cross-check whether the mapping from the upstream and downstream files (.github/sync-params.yaml, under planning section) is set correctly.

It looks the all mapping is set correctly.

This PR doen't update the paramters themselves and it just updates the comments; It does not affect the vehicle behavior.

PR #1867, opened 2026-06-09
feat(autoware_tensorrt_yolox): add HAZARD label and update remap

a-maumau:
## Description
This PR will add HAZARD class update in the `autoware_tensorrt_yolox`'s remap files.

This PR must merge with https://github.com/autowarefoundation/autoware_universe/pull/12730 .

## Notes for reviewers

None.

## Effects on system behavior

None.


autoware_launch/config/perception/traffic_light_recognition/tensorrt_yolox/roi_label_remap.csv
```diff
@@ -6,5 +6,6 @@ BUS, 3
 BICYCLE, 6
 MOTORBIKE, 5
 PEDESTRIAN, 7
-ANIMAL, 0
+ANIMAL, 8
+HAZARD, 9
 DEBRIS, 0
```
badai-nguyen: @a-maumau the same with unvierse side comment, currently all UNKNOWN could be treated as HAZARD, so shall we remap UNKNOWN and DEBRIS as HAZARD here? 
```suggestion
HAZARD, 9
DEBRIS, 9
```

technolojin (approved):

PR #1872, opened 2026-06-10
feat(perception): update tracker assignments to use general_vehicle_tracker for bounding box and polygon types

technolojin:
## Description
Replace multi_vehicle_tracker to general_vehicle_tracker

- multi_vehicle_tracker: double tracker that has normal vehicle tracker and git vehicle tracker. double work
- general_vehicle_tracker: single vehicle tracker having general parameters

Double tracker approach may taken to cover sudden size changes, when detector confused.
Sudden size change is absorbed by [UnstableShapeFilter](https://github.com/autowarefoundation/autoware_universe/pull/10864), so this kind of high-cost algorithm is not necessary anymore.

## How was this PR tested?

Tested by logging simulator, using sample rosbag file and TIER IV INTERNAL recordings.
This configuration is tested under [new PR](https://github.com/autowarefoundation/autoware_universe/pull/12728)

## Notes for reviewers

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #1877, opened 2026-06-11
ci(sync-params): add warning and how-to header for sync-params workflow, add signoff

paulsohn:
## Description

Due to the unfriendly interface of sync-params managed param files, developers are often confused which files in autoware_universe and autoware_core to update, resulting in manually updating params in autoware_launch, which is NOT an intended usage of the sync-params.

See https://github.com/autowarefoundation/autoware_launch/pull/1855#issuecomment-4622485030 .

image

The purpose of this workflow is to standardize auto-align schemas provided by component and schemas actually used in the integration (autoware_launch).

For example, the idiomatic procedure for adding ANIMAL / HAZARD related params ( https://github.com/autowarefoundation/autoware/issues/7137 ) is:

1. add ANIMAL and HAZARD labels in config files in autoware_universe first. If the implementation requires ANIMAL & HAZARD labels, then the example param file should contain them!
2. run sync-param workflow (or run `python .github/scripts/sync_params.py perception` to update the version.

image

3. override parameter AFTER sync-param has been applied, using `# {OVERRIDE}` marker comment.

After merging this PR, the new format will add the following header to sync-params managed files:

```
# DO NOT EDIT THIS FILE DIRECTLY FOR ADDING OR REMOVING FIELDS. Follow these steps:
#
# 1. Modify the param file in the upstream repository.
# 2. Run the sync-params workflow or run `python .github/scripts/sync_params.py perception` to update this file.
# 3. If local changes are needed, annotate them with an override marker (see below).
```

plus this adds signoff for workflow commit, addressing DCO failing issue.

Since perception params are broken now, green checks will be available for subsequent PRs.

## Notes for reviewers

None.

## Effects on system behavior

None.


badai-nguyen (approved):
I confirmed the chnage works on my local.
 `python3 .github/scripts/sync_params.py perception` genenrates the header and sync different params

```
7:~/autoware/src/launcher/autoware_launch$ git diff
diff --git a/autoware_launch/config/perception/object_recognition/detection/camera_vru_detection/near_range_camera_vru_detector.param.yaml b/autoware_launch/config/perception/object_recognition/detection/camera_vru_detection/near_range_camera_vru_detector.param.yaml
index c2b1e96e..febb90b2 100644
--- a/autoware_launch/config/perception/object_recognition/detection/camera_vru_detection/near_range_camera_vru_detector.param.yaml
+++ b/autoware_launch/config/perception/object_recognition/detection/camera_vru_detection/near_range_camera_vru_detector.param.yaml
@@ -1,4 +1,10 @@
 # This file is managed by sync-params workflow.
+# DO NOT EDIT THIS FILE DIRECTLY FOR ADDING OR REMOVING FIELDS. Follow these steps:
+#
+# 1. Modify the param file in the upstream repository.
+# 2. Run the sync-params workflow or run `python .github/scripts/sync_params.py perception` to update this file.
+# 3. If local changes are needed, annotate them with an override marker (see below).
+#
 # Keys marked with '{OVERRIDE}' or '# {OVERRIDE: reason}' are persisted as variant overrides.
 # 'reason' can be any single-line text without braces.
 # To keep local changes, annotate fields with '# {OVERRIDE}' or '# {OVERRIDE: reason}'.
diff --git a/autoware_launch/config/perception/object_recognition/detection/clustering/euclidean_cluster.param.yaml b/autoware_launch/config/perception/object_recognition/detection/clustering/euclidean_cluster.param.yaml
index 8c68a6ec..04595f71 100644

```

PR #1884, opened 2026-06-15
feat(euclidean_cluster): update cluster parameters for universe renames and per-label/confusable merge

technolojin:
## Description

Updates the euclidean-cluster launch configs to match the renamed and restructured clustering
parameters in autowarefoundation/autoware_universe#12738.

- **Parameter renames** (unit-explicit / clearer names) across `euclidean_cluster`,
  `voxel_grid_based_euclidean_cluster`, and `label_based_euclidean_cluster` configs:
  - `tolerance` → `tolerance_m`
  - `voxel_leaf_size` → `voxel_leaf_size_m`
  - `min_cluster_size` → `min_points_per_cluster`
  - `min_points_number_per_voxel` → `min_points_per_voxel`
  - `min_voxel_cluster_size_for_filtering` → `large_cluster_voxel_count_threshold`
  - `max_points_per_voxel_in_large_cluster` → `large_cluster_max_points_per_voxel`
  - `max_voxel_cluster_for_output` → `max_voxels_per_cluster`
- **Dropped `max_cluster_size`** from the voxel-grid and label-based configs — oversized clusters
  are now split into smaller sections instead of being dropped, so the parameter is no longer read.
- **Adopted the new `label_based` defaults** plus the new `label_cluster_params` per-label overrides
  (pedestrian / bicycle / hazard) and the `confusable_label_groups` block (truck_trailer).
- Mapped `tractor_unit` and `semi_trailer` to `trailer` so the truck/trailer confusable merge applies.

These renames are mandatory: the updated nodes read the new parameter names with no fallback, so the
launch configs must be updated in lockstep with the universe change.

## Related links

- Upstream universe PR: autowarefoundation/autoware_universe#12738
- [TIER IV INTERNAL](https://tier4.atlassian.net/browse/RDDR-397)

## How was this PR tested?

Parameter renames verified against the updated parameter declarations and JSON schemas in
autoware_universe#12738.

## Notes for reviewers

`config/.../clustering/euclidean_cluster.param.yaml` is a sync-managed file; its rename is applied
here so deployment stays consistent until the sync-params workflow re-syncs it after the universe PR
merges.

## Effects on system behavior

- Large clusters that were previously dropped are now split and emitted as smaller sections.
- `label_based` clustering now applies per-label parameters and post-merges confusable label pairs
  (truck/trailer), changing the number/labeling of objects for affected classes.


ktro2828 (approved):
LGTM

paulsohn:
euclidean_tracker: consider updating autoware_universe example param file as well.

image

there are many other universe - launch mismatches which should be handled in separate PRs.

technolojin:
@paulsohn 
I think you misunderstand how the parameter shall be updated during function development.

1. Update node package, who define and consume the parameters https://github.com/autowarefoundation/autoware_universe/pull/12738, not merge yet
2. Update the launcher parameter files (this PR) in the same time.
3. Merge 1 and 2 in the same time. Minimize launch-fail downtime.

With the check-params and CI sync method, it requires 1 and merge to main. With this approach, there is downtime until 2 is done by the CI sync method. 
The autoware has multiple contribution in the same time, this downtime can cause downtime all time long.

Please consider this sequence to your CI design.

paulsohn:
@technolojin 
With the introduction of this CI, that sequence itself should be updated like this:

1. update autoware_universe and its params
2. autoware_launch params can be **auto updated** with sync-param after the autoware_universe has been merged. No manual, parallel PR is required, and that "downtime" is just a few minutes.

(there are some param files not registered so I understand we need manual PRs for now, like in this PR)

3. create manual PRs only when override is needed, or sync-params does not work, or the launch file itself should be updated.

YoshiRi (approved):
LGTM

PR #1887, opened 2026-06-16
fix: update class_names in label_based_euclidean_cluster parameters for new ptv3 model

technolojin:
## Description

update class_names in label_based_euclidean_cluster parameters for new ptv3 model

model update: https://github.com/autowarefoundation/autoware/pull/7160

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/config/perception/object_recognition/detection/euclidean_cluster/label_based_euclidean_cluster.param.yaml
+++ b/autoware_launch/config/perception/object_recognition/detection/euclidean_cluster/label_based_euclidean_cluster.param.yaml
@@ -50,7 +50,7 @@
     # Semantic label selection and mapping
     # YAML declaration order is treated as the input class_id order.
     # cspell:ignore pullable
-    class_names:
+    class_names: # {OVERRIDE}
       car: car
       truck: truck
       bus: bus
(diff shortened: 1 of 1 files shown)


technolojin:
@paulsohn 

In this kind of structured parameter, how the parameter set shall be marked as OVERRIDE? Please see the file.

YoshiRi (approved):
LGTM for quick fix but we definitely need to quit this kind of class management.

autoware_launch/config/perception/object_recognition/detection/euclidean_cluster/label_based_euclidean_cluster.param.yaml
```diff
@@ -50,33 +50,22 @@
     # Semantic label selection and mapping
     # YAML declaration order is treated as the input class_id order.
     # cspell:ignore pullable
-    class_names:
-      drivable_surface: ignore
-      other_flat_surface: ignore
-      sidewalk: ignore
-      manmade: ignore
-      vegetation: ignore
+    class_names: # {OVERRIDE}
```
paulsohn, line 53:
```suggestion
    class_names:
```
Currently the override feature for sync-param only works for param files with a solid schema (without arbitrary key-value map) and `# {OVERRIDE}` only applies to parameter level. (i.e. a scalar or a list of scalars).

It doesn't support this "override everything under class_names parameter namespace" usage.

This file is not sync-params managed, so we don't need this marker.

technolojin: leave that mark for future sync

PR #1905, opened 2026-06-30
chore(rviz): adjust object recognition line widths and disable raw detection by default

technolojin:
## Description

Adjusts the object-recognition visualization in the RViz configs so the
perception output is easier to read, and changes the default detection toggle.

For every `Perception/ObjectRecognition` group across all RViz configs:

- **Detected objects** (`Detection` group): line width `0.10` → `0.03`, so the
  many raw per-detector boxes (Centerpoint, Stream PETR, PTv3 Cluster, Irregular
  object, PTv3 Object, BEV Fusion, …) render as thin outlines and overlap less.
- **Tracked / predicted objects**: line width `0.03` → `0.15`, so the final
  tracking/prediction output stands out clearly over the raw detections.
- **`Perception/ObjectRecognition/Detection` group**: `Enabled: true` → `false`,
  so the raw detection layer is hidden by default and only the
  tracking/prediction output is shown on startup.

Files updated:

- `autoware_launch/rviz/autoware.rviz`
- `autoware_launch/rviz/autoware_test_utils.rviz`
- `autoware_launch/rviz/perception.rviz`
- `autoware_launch/rviz/planning_bev.rviz`
- `autoware_launch/rviz/planning_tpv.rviz`
- `autoware_launch/rviz/scenario_simulator.rviz`

Changes are scoped strictly to the `ObjectRecognition` group; `DetectedObjects` /
`PredictedObjects` displays in other groups (planning debug, etc.) are untouched.

## How was this PR tested?

Visual check in RViz: detected-object boxes render as thin outlines, tracked /
predicted objects render with thicker outlines, and the raw `Detection` layer is
off by default and can still be re-enabled from the display tree.

## Notes for reviewers

Display-only RViz config change — no runtime/node behavior is affected. The
`Detection` group is only toggled off by default; it remains available to
enable manually.

## Effects on system behavior

None.

### DIFF
--- a/autoware_launch/rviz/autoware.rviz
+++ b/autoware_launch/rviz/autoware.rviz
@@ -835,7 +835,7 @@ Visualization Manager:
                     Yaw: false
                     Yaw Rate: false
                   Enabled: true
-                  Line Width: 0.10000000149011612
+                  Line Width: 0.03
                   Name: Centerpoint(red)
                   Namespaces:
                     {}
@@ -878,7 +878,7 @@ Visualization Manager:
                     Yaw: false
                     Yaw Rate: false
                   Enabled: true
-                  Line Width: 0.10000000149011612
+                  Line Width: 0.03
                   Name: Stream PETR(yellow)
                   Namespaces:
                     {}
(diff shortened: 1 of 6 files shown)


YoshiRi (approved):
LGTM

PR #1912, opened 2026-07-07
fix(minimum_rule_based_planner): update optimizer/modifier to processor

maxime-clem:
## Description

Update broken dependencies caused by merging https://github.com/autowarefoundation/autoware_universe/pull/12934 after https://github.com/autowarefoundation/autoware_universe/pull/12947

Also requires https://github.com/autowarefoundation/autoware_universe/pull/12954

## Notes for reviewers

None.

## Effects on system behavior

None.


Kotakku (approved):
~It looks like we also need to update the `exec_depend` tag for `autoware_trajectory_optimizer` in `autoware_launch/package.xml` to resolve the rosdep install dependencies.~
Sorry, I misread it. :pray:

zulfaqar-azmi-t4 (approved):

PR #1922, opened 2026-07-21
ci: distribute lint-autoware-system-design-format via sync-files

technolojin:
## Description

Bring the Autoware system design format lint workflow under `sync-files` management by adding `.github/workflows/lint-autoware-system-design-format.yaml` to the `.github/sync-files.yaml` mapping. It is distributed from [sync-file-templates](https://github.com/autowarefoundation/sync-file-templates) (source added in autowarefoundation/sync-file-templates#91). autoware_launch carries `*.node.yaml`, `*.module.yaml`, `*.system.yaml`, and `*.parameter_set.yaml` design files, so the workflow lints them on PRs.

The synced workflow file itself lands via the next `sync-files` run (nightly cron or manual `workflow_dispatch`).

## How was this PR tested?

CI-configuration change only. Verified the mapping entry matches the sync-file-templates source.

## Notes for reviewers

None.

## Effects on system behavior

None.


technolojin:
achieved by https://github.com/autowarefoundation/autoware_launch/pull/1955

PR #1930, opened 2026-07-29
refactor(tier4_perception_launch)!: select centerpoint variant folder in the launcher

technolojin:
## Description

Launcher and parameter-mirror side of the `lidar_centerpoint` model-bundle restructure. The bundle moves from a flat directory to one self-contained folder per variant (`base` / `tiny` / `sigma` / `short_range`) with uniform file names; the node-side change is in the companion `autoware_universe` PR.

**Variant selection moves into the launcher.** `lidar_dnn_detector.launch.xml` maps the model name to the variant folder (`centerpoint` → `base`, `centerpoint_<x>` → `<x>`) and passes the folder to `lidar_centerpoint.launch.xml` as `model_path`. The short_range block appends a static `/short_range` to `centerpoint_model_path`. Everything the dispatch used to compose by hand — `model_name`, `model_param_path`, `ml_package_param_path`, `class_remapper_param_path` — is gone; the manifest inside the variant folder owns those.

- Deleted the per-variant mirrors `lidar_model/centerpoint.param.yaml`, `centerpoint_tiny.param.yaml` and `centerpoint_sigma.param.yaml`, plus their `sync-params.yaml` entries. That layer no longer exists upstream.
- `lidar_model/centerpoint_common.param.yaml` stays mirrored and gains the node-implementation keys (`cloud_capacity`, `post_process_params`, `densification_params`) that moved out of the per-variant files.
- `centerpoint_common.param.yaml` is now wired as an explicit file-path argument, like the other general param files: `tier4_perception_component.launch.xml` sets `object_recognition_detection_centerpoint_common_param_path` → `perception.launch.xml` maps it to `centerpoint_common_param_path` → `detection.launch.xml` forwards it → both dispatch blocks pass it as `common_param_path`. The dispatch declares it with a pkg-share default so standalone use still works.
- Removed as dead interface: the `short_range_centerpoint_model_path` argument and the `lidar_short_range_detection_model_name` argument chain (`perception.launch.xml` → `detection.launch.xml` → `lidar_dnn_detector.launch.xml`). The short_range folder is derived from `centerpoint_model_path`, and the model name was never anything other than its own default.
- `sample_system_perception.parameter_set.yaml` updated to the manifest layout (`model_path` + `base/ml_package.param.yaml`).

## How was this PR tested?

- `pre-commit run --all-files`
- `colcon build`
- Launch smoke test against a local staging copy of the `v4.0` bundle, driven through `perception.launch.xml`: `base` and `tiny` ran the full path (engine build → inference → published objects); `sigma` and `short_range` were verified up to parameter load and manifest resolution.

Not yet done: engine-build smoke test for `sigma` and `short_range`. The test machine's GPU driver got wedged partway through and the run has not been repeated.

## Notes for reviewers

**This PR must land together with:**

- autowarefoundation/autoware_universe#13087 — node and package side of the same restructure
- autowarefoundation/autoware#7236 — ansible artifacts revision `v3.0` → `v4.0`

The `v3.0` bundle layout does not work with this code, and the `v4.0` layout does not work with the current code. Deployments pinned to `v3.0` data plus pre-merge code are unaffected.

Model bundle PR (draft, awaiting `v4.0` tag): <https://huggingface.co/AutowareFoundation/lidar_centerpoint/discussions/1>

The three deleted mirrors are removed from `sync-params.yaml` in the same commit, so the sync workflow will not try to recreate them.

## Effects on system behavior

No behavioral change for the `base` and `tiny` variants: the same weights and the same parameter values are loaded, only the file layout and the loading path differ.

`sigma` and `short_range` become usable out of the box. Their weights ship in the bundle for the first time — previously `short_range_centerpoint_model_path` pointed at a directory nothing populated, and `sigma` had no published ONNX at all.

Anyone overriding `lidar_short_range_detection_model_name` or `short_range_centerpoint_model_path` from an external launch file will get an unknown-argument error and should drop the override.


YoshiRi (approved):
LGTM

xmfcx (changes requested):
- https://github.com/technolojin/autoware_launch/pull/4

Created this to fix roscope drift, could you review?

xmfcx (approved):

autoware_launch/config/perception/object_recognition/detection/lidar_model/centerpoint_common.param.yaml
```diff
@@ -13,6 +13,15 @@
 
 /**:
   ros__parameters:
+    cloud_capacity: 2000000
```
mojomex, line 16: Presumably copied from somewhere, but 2'000'000 is way too much - our current beefiest architecture has around ~300'000.

tier4_universe_launch/tier4_perception_launch/launch/perception.launch.xml
```diff
@@ -292,7 +288,6 @@
           <arg name="lidar_detection_model_name" value="$(var lidar_detection_model_name)"/>
           <arg name="use_short_range_detection" value="$(var use_short_range_detection)"/>
           <arg name="lidar_short_range_detection_model_type" value="$(var lidar_short_range_detection_model_type)"/>
```
mojomex, line 290: Can these be removed as well? Not familiar with centerpoint but seems like this info would already be encoded in the short-range ML package.

tier4_universe_launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
```diff
@@ -287,9 +284,9 @@
       <arg name="lidar_detection_model_type" value="$(var lidar_detection_model_type)"/>
       <arg name="lidar_detection_model_name" value="$(var lidar_detection_model_name)"/>
       <arg name="lidar_short_range_detection_model_type" value="$(var lidar_short_range_detection_model_type)"/>
-      <arg name="lidar_short_range_detection_model_name" value="$(var lidar_short_range_detection_model_name)"/>
       <arg name="use_short_range_detection" value="$(var use_short_range_detection)"/>
       <arg name="output/short_range_objects" value="$(var lidar_ml_detector/output/short_range_objects)"/>
```
mojomex, line 288: Same as last comment.

mojomex (commented):
Not sure if I'm the right person to review this but I left some surface-level feedback.

tier4_universe_launch/tier4_perception_launch/launch/perception.launch.xml
technolojin: @mojomex the centerpoint shortrange is the only model that specified for short range, and as a secondary lidar ml model. but it dose not mean the lidar short range directly indicating the centerpoint variant....

mojomex:
I'm afraid I'm unqualified to review this - I have never touched Centerpoint nor its launchers so I can't judge its best launcher architecture...

technolojin:
@mojomex your approval is requested because of the system design file update. please review focus on the parameter-set file

mojomex (approved):
System design file changes LGTM!

Issue #1848, opened 2026-05-22
Introduce platform-specific launch packages


isamu-takagi:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I've agreed with the maintainers that I can plan this task.

### Description

Use platform-specific.launch.xml instead of launching autoware.launch.xml by providing vehicle_model and sensor_model as arguments, See https://github.com/orgs/autowarefoundation/discussions/7111 for details.

### Purpose

- Simplify the adaptation to new vehicles and sensors.
- Remove restrictions on package names and structures.

### Possible approaches

For the time being, autoware.launch.xml will support both the traditional direct launch mode and the platform_launch package mode.

Autoware Core

1. Remove vehicle_model and sensor_model arguments from all files except the top-level launch files. 
    - https://github.com/autowarefoundation/autoware_core
2. Create sample_platform_launch by merging sample description and launch packages.
    - https://github.com/autowarefoundation/autoware_core
3. Update Autoware Documentation
    - https://github.com/autowarefoundation/autoware-documentation

Autoware Universe 

1. Remove vehicle_model and sensor_model arguments from all files except the top-level launch files. 
    - https://github.com/autowarefoundation/autoware_launch
    - https://github.com/autowarefoundation/autoware_universe
2. Create other platform_launch packages for demos. 
    - https://github.com/autowarefoundation/autoware_launch
3. Update Autoware Documentation
    - https://github.com/autowarefoundation/autoware-documentation

### Definition of done

- [ ] Remove model arguments and create sample_platform_launch
    - [ ] https://github.com/autowarefoundation/autoware_core/pull/1102
- [ ] Create other platform_launch packages for demos
- [ ] Update Autoware Documentation

isamu-takagi:
It must be capable of being used across both actual vehicles and various simulation methods. https://github.com/autowarefoundation/autoware_launch/issues/1848

PR #1938, opened 2026-08-11
refactor(sensing)!: separate sensor driver and pre-processing layers from sensor kit configuration

technolojin:
## Description

This PR separates the sensing launch stack into layers by **what varies**, so that the parts independent of any
particular system design become reusable packages, and the parts encoding one concrete system stay in the kit.

| Layer | Package | Role | Varies with |
|---|---|---|---|
| Hardware interface | `autoware_sensor_driver_launch` | brings a physical sensor into the ROS graph — per-make driver chain, per-model wrapper | sensor make/model only |
| Data pre-processing | `autoware_sensing_launch` | reusable units — container, crop boxes, distortion correction, outlier filtering, concatenation, IMU/GNSS/radar post-processing, blockage diagnostics | nothing; pure pipeline units |
| System configuration | `*_sensor_kit_launch` | one concrete sensor system: which sensors exist, where they sit, which containers they run in, which units they are wired through, with which parameters | the specific system design |

**The sensor kit is still called from `autoware_launch`** through `$(var sensor_model)_launch`, exactly as
before. Changing that entry point is a separate step, deliberately not part of this PR.

## Ownership contract

**Sensor kits create containers. Units only load nodes into them.** `pointcloud_container.launch.xml` is the
single unit that creates a container, and it loads nothing into it. Every other unit takes `target_container`,
the **absolute path** of an existing container, and never derives it from its own namespace —
`load_composable_node` ignores `push-ros-namespace`. The caller passes a literal matching its own namespace
pushes; the two must be kept in step. The concatenation units are the exception, keeping the historical
`pointcloud_container_name` passed through from the top level. This is what makes a driver reusable: a platform
package creates its container and loads driver and pre-processing into it, knowing how neither is built.

## What changed

**1. The `tier4_sensing_launch` layer is gone.** It was a 19-line pass-through that pushed the `sensing`
namespace and included the kit's `sensing.launch.xml` — no nodes, no configuration.
`component_sensing.launch.xml` (renamed from `tier4_sensing_component.launch.xml`) now includes the kit entry
directly and pushes the namespace itself. The package is renamed `autoware_sensing_launch` and moved under a new
top-level `autoware_universe_launch/`.

**2. Drivers are split from data processing.** Driver wiring, container creation and pre-processing topology were
entangled in one Python generator, so none of the three could be reused or changed on its own. Neither new layer
depends on the other; kits declare both as `<exec_depend>`, so the sensing pipeline is rosdep-resolvable for the
first time. The forked `common_sensor_launch`, `single_lidar_common_launch` and `common_awsim_labs_sensor_launch`
are removed.

**3. All six Python launch generators (1227 lines) become declarative XML.**

| Removed | Replaced by |
|---|---|
| `common_sensor_launch/…/nebula_node_container.launch.py` (313) | `nebula_velodyne`, `nebula_robosense` chains |
| `single_lidar_common_launch/…/nebula_node_container.launch.py` (307) | the same shared chains |
| `common_awsim_labs_sensor_launch/…/velodyne_node_container.launch.py` (227) | `pointcloud_container` + `pointcloud_preprocessor_cpu` |
| `{sample,awsim_labs}_sensor_kit_launch/…/pointcloud_preprocessor.launch.py` (99, 98) | `pointcloud_concatenation_cpu` |
| `carla_sensor_kit_launch/…/pointcloud_preprocessor.launch.py` (183) | `pointcloud_crop_boxes` |

The two `nebula_node_container.launch.py` files were near-identical copies that had drifted: the `single_lidar`
fork had lost the container namespace handling, `udp_only` support and the pre-processing diagnostic thresholds
the sample version had gained. Collapsing them to one chain per make removes the fork and the drift together.
Per-model parameters now come from the model param file shipped in the nebula driver package, and crop box
geometry from `generate_crop_box_params.py <vehicle_info> <mirror> <out_dir>`, committed to the kit that owns the
vehicle. No Python launch files remain in the sensing path. Net: +1200 / −2121 lines.

## Provenance of newly added files

Most driver files are relocated from `common_sensor_launch` (`velodyne_{VLP16,VLS128}`,
`robosense_{Bpearl,Helios}`), and the `nebula_velodyne` / `nebula_robosense` chains derive from the removed
`nebula_node_container.launch.py`. The genuinely new content is **ported from the public
[tier4/aip_launcher](https://github.com/tier4/aip_launcher)**, package `aip_common_sensor_launch`: the
`hesai_{XT32,OT128}` wrappers (put on the per-make chain pattern — `target_container` and crop box param paths
added, `use_shared_container` / `use_cuda_preprocessor` dropped, `scan_phase` → `sync_angle` + `cut_angle`,
`host_ip` default from broadcast to `192.168.1.10`), `septentrio_heading_converter.py` (identical but for two
topic names made relative so the unit is namespace-portable), and the blockage and radar parameter files
(byte-identical). `nebula_hesai.launch.xml` and the remaining new units have no counterpart upstream —
`aip_launcher` launches those nodes from inside its Python container generator — and follow the same pattern
here. Both repositories are Apache-2.0 with TIER IV copyright; existing headers are retained.

## Relation to the platform launch proposal

Groundwork for [discussion #7111](https://github.com/orgs/autowarefoundation/discussions/7111) /
[issue #1848](https://github.com/autowarefoundation/autoware_launch/issues/1848), which ask for a
`*_platform_launch` package launching the sensing drivers **directly**. Today that means reproducing the Python
container generator per platform package; after this PR it is two includes — `pointcloud_container`, then the
per-model wrapper with `target_container`. The missing `package.xml` depends and the runtime `mirror.param.yaml`
lookup are addressed as a side effect (`vehicle_mirror_param_file` is still plumbed for interface compatibility
but no longer read). Inverting the entry so the kit `sensing.launch.xml` disappears as a contract is the next
step, and the one this PR unblocks.

## How was this PR tested?

**Running-system A/B, `main` (`da34f6f35`) vs this branch (`f63a83223`), every kit.** `ros2 launch --print` is not
usable as evidence on Humble — it prints top-level action objects and never expands the includes — so each side was
launched for real: `logging_simulator.launch.xml` with `sensing:=true vehicle:=true` and every other subsystem
`false`, on an isolated `ROS_DOMAIN_ID`, 25 s settle. Captured per kit: `ros2 node list`, `ros2 component list`,
`ros2 topic list`, `ps` container arguments (executable + name + namespace), and `ros2 param dump` of every
`/sensing` node. Each side was built from a pruned `install/` so packages deleted on the other side cannot resolve.

| kit / vehicle | sensing nodes | components | container processes | topics | param dumps |
|---|---|---|---|---|---|
| `sample_sensor_kit` / `sample_vehicle` | 29 = 29 | identical | identical | 175 = 175 | 29 identical, 0 differing |
| `awsim_sensor_kit` / `sample_vehicle` | 10 = 10 | identical | identical | 132 = 132 | 10 identical, 0 differing |
| `awsim_labs_sensor_kit` / `awsim_labs_vehicle` | 19 = 19 | identical | identical | 149 = 149 | 19 identical, 0 differing |
| `carla_sensor_kit` / `sample_vehicle` | 6 = 6 | identical | identical | 123 = 123 | 6 identical, 0 differing |
| `single_lidar_sensor_kit` / `sample_vehicle` | 4 → 7 | 1 container renamed | 1 namespace changed | 116 → 124 | see below |

Counts exclude anonymous `transform_listener_impl_*` entries. Container executables (`component_container` vs
`component_container_mt`), names and namespaces match exactly on the four unchanged kits — including the per-LiDAR
`use_multithread` split, which is now a kit call-site value. The crop box migration produces no parameter
difference: the committed files reproduce what the removed Python generator computed.

`single_lidar_sensor_kit` is the one kit whose graph changes, because it is broken on `main`: three of its four
LiDAR composable nodes fail to construct there.

```
Failed to load 'velodyne_ros_wrapper_node' : Statically typed parameter 'udp_only' must be initialized.
Failed to load 'distortion_corrector_node': Statically typed parameter 'processing_time_threshold_sec' must be initialized.
Failed to load 'ring_outlier_filter'      : Statically typed parameter 'processing_time_threshold_sec' must be initialized.
```

`single_lidar_common_launch` had drifted from `common_sensor_launch`: its `nebula_node_container.launch.py` never
passed `udp_only`, and its private `distortion_corrector_node` / `ring_outlier_filter_node` parameter files lack
`processing_time_threshold_sec` (and `timestamp_mismatch_fraction_threshold`), which the current
`autoware_pointcloud_preprocessor` requires. On this branch the kit uses the shared unit configs and the driver
layer, all four nodes load, and the launch log contains no ERROR. `main`'s `awsim_sensor_kit` log is the only other
one with errors — the nebula driver that never reached the graph, see below.

Also: `rosdep` resolves through the new `exec_depend` entries; `colcon build` over all touched packages;
`pre-commit` clean at every commit.

## Notes for reviewers

Seven commits, each independently reviewable and buildable; reviewing in order is recommended. The two `feat`
commits are purely additive — nothing includes the new units yet — so the four kit commits are where behavior can
change, one kit at a time.

| Commit | Size |
|---|---|
| `88141c22` rename `tier4_sensing_launch` to `autoware_sensing_launch` | 13 files, +35/−307 |
| `a100c16f` add the shared sensing unit library | 20 files, +591/−4 |
| `6154c3d5` add the sensor driver layer | 13 files, +446/−0 |
| `7f56c2f9` adopt the units in the sample and awsim kits | 23 files, +142/−687 |
| `a13faa0c` adopt the units in the single_lidar kit | 14 files, +40/−561 |
| `6ebca6df` adopt the units in the awsim_labs kit | 14 files, +63/−529 |
| `f63a8322` adopt the units in the carla kit | 7 files, +50/−200 |

Where to look hardest: in `a100c16f` and `6154c3d5`, compare node set, parameter sources and remappings of the
removed generator against the new XML chain for one LiDAR (sample / `top` / VLS128), and check the ported Hesai
wrappers against their `aip_launcher` originals. In the four kit commits, each kit's `container_namespace` /
`target_container` literal must match its own `push-ros-namespace` chain — only a running graph reveals a
mismatch.
`use_multithread` is a kit call-site value now too: VLS128 and hesai want `true`, VLP16 and robosense `false`, or
the container silently downgrades to single-threaded.

### Known gaps

- Six units have no consumer here: `pointcloud_{preprocessor,concatenation}_cuda`, `blockage_diagnostics`,
  `gnss_postprocessor`, `imu_corrector`, `radar_tracks_postprocessor`. They are what a platform package needs,
  ported so the library is complete, but nothing exercises them. The kits still hand-roll the imu corrector and
  gyro bias estimator in their own `imu.launch.xml`; including `imu_corrector.launch.xml` there would put a unit
  under test.
- `sensor_model` is overloaded: inside `autoware_sensor_driver_launch` it means the device model (`VLS128`,
  `OT128`), unrelated to the top-level kit argument. `device_model` would be a better name.
- The `target_container` literals duplicate each kit's namespace pushes.

## Effects on system behavior

The node graph is unchanged for the sample, awsim, awsim_labs and carla kits, as measured above. Three deliberate
exceptions, each stated in the commit that makes it:

- **`single_lidar_sensor_kit_launch`** gains `udp_only` support and the pre-processing diagnostic thresholds its
  removed fork had lost, so the driver, distortion corrector and ring outlier filter now load instead of throwing
  at construction. Its per-LiDAR container also moves from
  `/sensing/lidar/top/autoware_pointcloud_preprocessor/pointcloud_container` to
  `…/top/pointcloud_preprocessor/pointcloud_container`, matching every other kit.
- **`awsim_sensor_kit_launch`** no longer includes a driver. AWSIM publishes `pointcloud_raw_ex` directly, and the
  driver node it did include never reached the graph — nebula rejects the kit's broadcast `host_ip` and the
  constructor throws. The include existed only to obtain a container, which the kit now creates itself, so the
  graph is unchanged and one startup ERROR disappears. The velodyne monitor goes with it, and
  `use_distortion_corrector` is dropped, since the removed generator ignored it.
- Driver `host_ip` defaults change from broadcast to nebula's own default, because current nebula rejects a
  broadcast address at construction even when `launch_hw` is false.

Crop box geometry no longer follows `vehicle_model` at runtime; it comes from the file the kit commits. Every
vehicle description here carries identical geometry, so nothing changes, but a kit paired with a
differently-shaped vehicle must regenerate its files.

**Breaking for downstream consumers:** `tier4_sensing_launch` and the three `common_*_launch` packages no longer
exist. Per-model wrapper filenames and argument names are unchanged except around the container: they no longer
accept `container_namespace`, `container_name`, `use_multithread`, or the crop box, distortion and ring outlier
arguments, and instead take `target_container`, which must already exist. Out-of-tree kits — `aip_launcher` among
them — have to create their container, include the pre-processing unit themselves, and update the package name.


badai-nguyen (approved):
@technolojin 
Thank you for your great work, 

Could you fix the conflict and rebase on the latest version. without PR (https://github.com/autowarefoundation/autoware_launch/pull/1930), error related to centerpoint will occurred. 

After rebased on[ the above PR](https://github.com/autowarefoundation/autoware_launch/pull/1930), 
I confirmed that after this refactoring, `logging_simulator.launch.xml`  and `e2e_simulator.launch.xml` could be run successfully. (Note

Image

yukkysaito (approved):
LGTM

badai-nguyen (approved):

YoshiRi (approved):
LGTM

PR #1942, opened 2026-08-13
refactor(perception)!: modularize perception launchers into reusable feature units

technolojin:
## Description

Part of #1939 (perception item), supporting
[discussion #7111](https://github.com/orgs/autowarefoundation/discussions/7111) and #1848.

Restructures the perception launch stack into modular, reusable feature launchers so that each
platform (sensor kit) can configure its own set of features and node groups.

### What was done

1. **Package rename**: `tier4_perception_launch` → `autoware_universe_launch/autoware_perception_launch`.
   The perception parameter tree moves from `autoware_launch` into the package, and feature launchers
   default their own parameter paths.
2. **Feature launchers**: perception is decomposed into self-contained feature launchers grouped by
   function — `object_recognition/` (detector / filter / merger units and per-modality compositions),
   `obstacle_segmentation/` (per-method ground segmentation), `occupancy_grid_map/`, and
   `traffic_light_recognition/` (per-scope detector / recognition / fusion units). Each launcher
   declares its interface (arguments, input/output topics) explicitly at the top of the file.
3. **Python → XML**: all remaining Python perception launchers are flattened to XML for readability
   (ground segmentation, pointcloud map filter, irregular object detector, traffic light launchers).
4. **Junction arguments removed**: mode-switch arguments that multiplied into untestable combinations
   are removed. Object detection reduces to five explicit compositions: `lidar`, `lidar_radar`,
   `camera`, `camera_lidar_radar`, and `camera_lidar_radar_serial`. Some switches remain for
   flexibility at this stage.
5. **System components**: `autoware_launch` composes the features in
   `components/component_perception.launch.xml`, with traffic light recognition split into its own
   `components/component_traffic_light.launch.xml`.

Developers of a new platform can compose these feature launchers — including mid-stage units such as
a single detector, filter, or recognition stage — on their own demand.

### Kept at this stage

- The current `autoware.launch.xml` entry trigger is unchanged.
- Topic namespaces are unchanged.
- The planning component still reads `compare_map_filter_param_path` from the `autoware_launch` copy
  (`config/perception/.../pointcloud_map_filter.param.yaml`); retargeting it to the
  `autoware_perception_launch` copy is deferred to a follow-up PR.

## How was this PR tested?

- E2E simulation (AWSIM).
- Planning simulation.
- Logging simulation with the sample rosbag.

## Notes for reviewers

Sibling PR of #1938 (sensing). Reviewing commit by commit follows the restructuring steps.

[...]

Issue #1939, opened 2026-08-12
Modularization of components


technolojin:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I've agreed with the maintainers that I can plan this task.

### Description

Refactoring and restructuring component launchers to following sections

- node launcher : thin wrapper of node launcher, mostly stored in node packages
- feature launcher : series of node launchers that make a feature. reusable unit
- system component launcher : implementation of feature launchers, combined for specific system

### Purpose

Improve the Autoware's modularity and re-usability.
Support https://github.com/orgs/autowarefoundation/discussions/7111 / https://github.com/autowarefoundation/autoware_launch/issues/1848

### Possible approaches

- Step 1 : Properly refactoring launchers.
- Step 2 : Apply Autoware system designer

### Definition of done

restructuring tier4_<component>_launch packages
- define features and store them to autoware_<component>_launch
- apply system specific launchers to sensor_kit platforms and autoware_launch
- remove tier4_<component>_launch packages

### Items

- sensing https://github.com/autowarefoundation/autoware_launch/pull/1938
- perception
  - migration https://github.com/autowarefoundation/autoware_launch/pull/1942
  - config as separated package https://github.com/autowarefoundation/autoware_launch/pull/1955
  - module design update https://github.com/autowarefoundation/autoware_launch/pull/1957
- planning 
  - migration https://github.com/autowarefoundation/autoware_launch/pull/1954
  - module design update https://github.com/autowarefoundation/autoware_launch/pull/1967
- control 
  - migration https://github.com/autowarefoundation/autoware_launch/pull/1962
  - modularization https://github.com/autowarefoundation/autoware_launch/pull/1965
- map
  - implement system designer https://github.com/autowarefoundation/autoware_launch/pull/1970 
  - modularization
- localization
  - implement system designer https://github.com/autowarefoundation/autoware_launch/pull/1968 
  - modularization https://github.com/autowarefoundation/autoware_launch/pull/1971
- system
- vehicle
- simulator
- (autoware_api)

PR #892, opened 2024-02-28
chore: add codeowners-manual

YamatoAndo:
## Description

I added a file to set the code owners to ensure that the responsible person for that file reviews it properly.

I've selected the code owners arbitrarily, but if there are any additions/removals of code owners, please change it individually after the merge of this PR.

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [ ] I've confirmed the [contribution guidelines].
- [ ] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #1954, opened 2026-08-21
refactor(planning)!: migrate tier4_planning_launch to autoware_planning_launch with a separate config package

technolojin:
## Description

Part of #1939 (planning item), supporting

Moves the planning launch stack out of the product layer and splits its parameter tree into a
dedicated package, so that a platform can supply its own planning parameters without forking the
launchers.

This is the migration step only: the planning launch graph itself
(`planning` → `scenario_planning` → `lane_driving` / `parking` → behavior / motion) is unchanged.
Decomposition into reusable feature units is left for a follow-up.

### What was done

1. **Package rename**: `tier4_planning_launch` → `autoware_universe_launch/autoware_planning_launch`,
   and the old package is removed.
2. **Separate config package**: the planning parameter tree moves from
   `autoware_launch/config/planning/` into a new data-only package
   `autoware_universe_launch/autoware_planning_config`, dropping the redundant `planning/` level.
   Unlike sensing and perception, the tree lives in its own package rather than inside the launch
   package, so a platform can depend on the launchers and the parameters independently.
3. **Config addressed by argument**: `planning.launch.xml` resolves the whole tree through
   `planning_config_pkg` (default `autoware_planning_config`). Setting it once at the entry point
   relocates every parameter file, since the launch configuration propagates through every include.
   Platforms that differ in only a few files instead override the individual `*_param_path`
   arguments, which are now all declared at the top of `planning.launch.xml` with defaults.
4. **Module preset moved in**: `planning.launch.xml` includes the preset itself and takes
   `module_preset`. Values passed in by a caller still win, since the include is the first statement.
5. **Thin system component**: `components/tier4_planning_component.launch.xml` (149 lines, ~70 of
   them parameter paths) becomes `components/component_planning.launch.xml` (36 lines) carrying only
   junctions, boundary topics and the global parameter loader. `planning_setting` is now passed
   explicitly rather than relying on it being inherited.
6. **Dead plumbing removed**: `compare_map_filter_param_path` and `pointcloud_container_name` were
   passed into planning but read by nothing. Dropping the former retires the last reference to
   `autoware_launch/config/perception/.../pointcloud_map_filter.param.yaml`, so that orphaned copy is
   deleted here — this closes the follow-up declared in #1942 (a plain deletion rather than the
   retarget originally planned).
7. **Configurable tracked objects topic**: the diffusion planner's tracked objects topic was
   hardcoded; it is now `input_tracked_objects_topic_name`, plumbed like the objects and pointcloud
   topics.

`planning.launch.xml` can now be launched standalone with no arguments, which was not possible before.

### Breaking changes

Out-of-tree callers must update:

- `tier4_planning_launch` → `autoware_planning_launch`
- `autoware_launch/config/planning/**` → `autoware_planning_config/config/**` (no `planning/` level)
- `components/tier4_planning_component.launch.xml` → `components/component_planning.launch.xml`
- `compare_map_filter_param_path` and `pointcloud_container_name` are no longer accepted

### Kept at this stage

- The launch graph, node names, namespaces and topics are unchanged.
- `behavior_planning_deprecated.launch.xml` is moved as-is; it is unreferenced but left alone.
- `Tier4PlanningWrapper.node.yaml` keeps its name (it is not referenced by any system design, which
  uses `Planning.module`); only its package name and parameter paths are updated.

## How was this PR tested?

- Equivalence of the expanded launch tree against `main`, for the default, `diffusion_planner` and
  `path_generator` variants: the node set, container set, composable node set and every resolved
  parameter file path are **byte-identical** in all three. All 62 composable node parameter files
  resolve out of `autoware_planning_config`.
- `planning.launch.xml` expands standalone with zero unresolved arguments and zero missing files.
- End-to-end propagation of the new tracked objects argument verified from the component down to the
  `autoware_diffusion_planner` include, with and without an override.
- `colcon build` and `colcon test` on `autoware_planning_launch`, `autoware_planning_config`,
  `autoware_launch` and `autoware_sample_designs`: 122 tests, 0 failures.
- `pre-commit run` clean.
- Planning simulator launched and manually triggered engage mode.

## Notes for reviewers

Sibling PR of #1938 (sensing) and #1942 (perception), but deliberately narrower — this one does not
restructure the launch graph, so the diff outside the two planning packages is mechanical.

| Area | Files | What to review |
|---|---|---|
| Planning launch package | `autoware_universe_launch/autoware_planning_launch/**`, `tier4_universe_launch/tier4_planning_launch/**` (removed) | the `planning_config_pkg` argument, the preset include position, and the parameter path block at the top of `planning.launch.xml` |
| Planning config package | `autoware_universe_launch/autoware_planning_config/**` | package boundary and the `config/` layout that overriding `planning_config_pkg` commits a platform to |
| Autoware entry | `launch/autoware.launch.xml`, `launch/components/component_planning.launch.xml` | what stays a product-level choice, and the new `<group>` scoping |
| Perception cleanup | `autoware_launch/config/perception/**` (removed) | that the deleted parameter copy really had no consumer |
| Repo metadata | `.github/CODEOWNERS`, `.github/CODEOWNERS-manual`, `.github/sync-params.yaml`, `autoware_launch/test/autoware_guideline_check.yaml` | 26 retargeted sync paths and the new config package ownership rules |
| Sample designs | `autoware_sample_designs/design/**` | retargeted parameter paths and the two dropped dead parameters |

🤖 Generated with [Claude Code](https://claude.com/claude-code)


technolojin:
Suggestion to code owners

* separate `planning.launch.xml` to modes (rule_based, diffusion_planner)
* combination of common + rule_based/diffusion_planner

.github/CODEOWNERS
```diff
@@ -2,6 +2,8 @@
 autoware_launch/** mfc@leodrive.ai ryohsuke.mitsudome@tier4.jp yukihiro.saito@tier4.jp
 autoware_sample_designs/** max.schmeller@tier4.jp maxime.clement@tier4.jp taekjin.lee@tier4.jp
 autoware_universe_launch/autoware_perception_launch/** dai.nguyen@tier4.jp masato.saeki@tier4.jp taekjin.lee@tier4.jp yoshi.ri@tier4.jp yukihiro.saito@tier4.jp
+autoware_universe_launch/autoware_planning_config/** fumiya.watanabe@tier4.jp kosuke.takeuchi@tier4.jp kyoichi.sugahara@tier4.jp mamoru.sobue@tier4.jp satoshi.ota@tier4.jp shumpei.wakabayashi@tier4.jp takayuki.murooka@tier4.jp zulfaqar.azmi@tier4.jp
```
shmpwk:
```suggestion
autoware_universe_launch/autoware_planning_config/** fumiya.watanabe@tier4.jp kosuke.takeuchi@tier4.jp kyoichi.sugahara@tier4.jp mamoru.sobue@tier4.jp shumpei.wakabayashi@tier4.jp takayuki.murooka@tier4.jp zulfaqar.azmi@tier4.jp
```

Ota-san will leave soon

autoware_launch/config/perception/object_recognition/detection/pointcloud_filter/pointcloud_map_filter.param.yaml
```diff
@@ -1,29 +0,0 @@
-/**:
```
shmpwk, line 1: Why is this removed in THIS planning launch refactoring

technolojin: see https://github.com/autowarefoundation/autoware_launch/pull/1942#issuecomment-5310827377

this is known that the file had dead link of the planning launch after https://github.com/autowarefoundation/autoware_universe/pull/11320

this file shall not be exist for planning.

.github/CODEOWNERS
technolojin: :sob:

technolojin: also sugahara-san

.github/CODEOWNERS
maxime-clem, line 1: As far as I understand `CODEOWNERS` file should not be edited manually.

.github/CODEOWNERS-manual
maxime-clem, line 1: As far as I understand `CODEOWNERS` file should not be edited manually.

.github/CODEOWNERS
technolojin: more fix in https://github.com/autowarefoundation/autoware_launch/pull/1954/changes/b929273a81d90931412d3c7d8479ce371dc37719

control component will be done in a following PR

shmpwk:
@technolojin 
Thank you for your reply. I have an another question about design.
Why a separate autoware_planning_config when sensing/perception colocate config/ in the launch package? exec_depend on the default config also means launch and config are not fully independent. How about aligning with the colocation pattern? Or could you document it?

technolojin:
@shmpwk 

> Why a separate autoware_planning_config

This is my proposal of the separation of common modular part and system(platform) specific part of the planning component.
The planning component dose not have large dependency of system (sensor set, control unit, other HW set), but it vary onto operational design domain (ODD), and it is adapted by configs/parameters. 

For example, the planning component has `module_preset` to switch on/off a lot of switch-arguments as set. This kind of argument makes launcher/parameter structure complex. It shall be flattened and simplified.

Therefore, I think the configuration shall be separated as a package. 
I would like to listen the component owners opinion. 

For next stage, https://github.com/orgs/autowarefoundation/discussions/7111 shall be implemented by platform specific planning configs having its own ODD-specific configurations.

> when sensing/perception colocate config/ in the launch package?

I am working on separation of config package for perception https://github.com/autowarefoundation/autoware_launch/pull/1955 (sensing will follow). It has a bit different requirements, and also to be align between components.

>  exec_depend on the default config also means launch and config are not fully independent

It would be better not to depends, but the autoware_planning_config is as default. the config package is lightweight (no more packages depends).

shmpwk (approved):
Thank you for the clarification. That makes sense.
No further objection from my side on this design point. Looking forward to the platform/ODD-specific config packages.

ogrex (approved):
LGTM

maxime-clem (approved):

.github/CODEOWNERS
technolojin: fixed https://github.com/autowarefoundation/autoware_launch/pull/1954/changes/ca24eb2eb1589c72a5031542981b1faf6c1e6340

.github/CODEOWNERS-manual
technolojin: by https://github.com/autowarefoundation/autoware_launch/pull/892, this is manually maintained file.

takayuki5168 (approved):

mitsudome-r (approved):

PR #1955, opened 2026-08-21
refactor(perception)!: extract the perception parameter tree into autoware_perception_config and modularize the simulated perception pipeline

technolojin:
## Description

Part of #1939 (perception and simulator items), supporting
Follow-up to #1942, which left the perception parameter tree inside the launch package.

1. **Separate config package**: the 63 parameter/CSV files under
   `autoware_perception_launch/config/` move verbatim into a new data-only package
   `autoware_universe_launch/autoware_perception_config`, so a platform can adopt the launchers and
   the parameters independently.
2. **Config addressed by argument**: every feature launcher declares `perception_config_pkg`
   (default `autoware_perception_config`) and resolves it once into a local `perception_config_dir`.
   `component_perception`, `component_traffic_light` and `planning_simulator/perception_simulation`
   each declare it, so a platform switches the whole tree at the entry point it launches; platforms
   differing in a few files keep overriding the individual `*_param_path` arguments.
3. **Simulated perception belongs to the product, not the simulator**: the simulated pipeline —
   dummy perception, dummy traffic light publisher, laserscan occupancy grid map, tracking,
   prediction, empty-objects stub, online evaluator / analytics publisher, elevation map loader —
   lives in `autoware_launch/launch/planning_simulator/perception_simulation.launch.xml`, included
   by `planning_simulator.launch.xml` as a sibling of the simulator component and launchable
   standalone. The occupancy grid map's implicit `unless scenario_simulation` guard becomes an
   explicit `launch_occupancy_grid_map` argument computed by the caller.
4. **Simulator freed of perception**: `simulator.launch.xml` keeps vehicle model, localization,
   doors, scenario adapter and fault injection only; fault injection is gated solely by
   `launch_fault_injection`, now passed explicitly through `tier4_simulator_component.launch.xml`
   instead of leaking through include scope. The perception `exec_depend`s move from
   `tier4_simulator_launch` to `autoware_launch`.
5. `.github/sync-params.yaml` and the `sync_params.py` docstring are retargeted. The manual
   `.github/CODEOWNERS-manual` entry for `autoware_perception_launch/config/**` is removed, so
   ownership follows `autoware_perception_config/package.xml` through the generator — which drops
   `lei.gu@tier4.jp` unless that address is added there. `autoware_sample_designs` now depends on
   `autoware_launch`, matching where all 219 of its parameter references resolve.

### Breaking changes

- `autoware_perception_launch/config/**` → `autoware_perception_config/config/**`, identical layout.
- `simulator.launch.xml` no longer launches simulated perception and drops every perception-related
  argument (`launch_simulator_perception_modules`, `launch_dummy_perception`, `perception/*`,
  `launch_dummy_traffic_light_publisher`, `dummy_traffic_light_mode`, `external/traffic_signals`,
  `sensing/visible_range`, `occupancy_grid_map_updater`, `use_pointcloud_container`,
  `use_perception_online_evaluator`, `use_perception_analytics_publisher`, `data_path`) and the
  parameter paths they carried.
- `tier4_simulator_component.launch.xml` requires `launch_fault_injection` (no default) and drops
  the same ten perception arguments. A caller launching this file directly must pass
  `launch_fault_injection` — a missing required argument is a hard launch error — and must include
  `perception_simulation.launch.xml` itself to keep simulated perception; the removed arguments
  become undeclared and are silently ignored.
- `planning_simulator.launch.xml` drops `perception/enable_traffic_light`. Setting it `true`
  previously selected `traffic_light_recognition_merge` on V2X `external/traffic_signals` and
  suppressed the dummy publisher, which now runs unconditionally.
  `traffic_light_recognition_merge.launch.xml` is left in place but is now unreferenced — whether it
  stays as a reusable page is a reviewer's call.

At `planning_simulator.launch.xml` defaults the launch graph, node names, namespaces and topics are
unchanged, and the parameter files are moved byte-for-byte.

## How was this PR tested?

- Launch-tree equivalence against `main` by read-only expansion: **40 perception entry points and
  variants** (`component_perception` across detection model, OGM, ground segmentation and filter
  options, `component_traffic_light` in all three topologies, all 22 feature launchers standalone)
  and **6 simulator configurations** — node set, container set, composable node set and every
  resolved parameter path identical once the config package name is accounted for. A
  `perception_config_pkg` override relocates all 20 parameter lookups of the two components.
- A real `planning_simulator.launch.xml` run (sample map, sample vehicle/sensor kit) starts dummy
  perception, dummy traffic light, tracker, prediction and analytics publisher, and loads the
  laserscan occupancy grid nodes into `pointcloud_container`, with no launch errors;
  `perception/enable_elevation_map:=true` additionally starts `elevation_map_loader_node`.
- `colcon build` + `colcon test` on the affected packages: 77 tests, 0 failures. `pre-commit` clean.

## Notes for reviewers

Sibling of #1938 (sensing), #1942 (perception) and #1954 (planning); the config split follows #1954,
so perception and planning end up with the same interface. Most of the diff is the 63-file move plus
the `find-pkg-share` → `perception_config_dir` rewrite;
`planning_simulator/perception_simulation.launch.xml` is the only genuinely new launch file to read.


takayuki5168 (approved):

badai-nguyen (approved):
@technolojin 
- It looks like there is a dangling pointcloud processing pipeline in the default perception mode: 

Image

  - `voxel_grid_downsample_filter → voxel_based_compare_map_filter → euclidean_cluster → shape_estimation → detected_object_feature_remover`

  - The output of detected_object_feature_remover does not appear to be consumed by any downstream node.
  - When use_semantic_segmentation_ptv3 is false (unless="$(var use_semantic_segmentation_ptv3)"), either the input of multi_object_tracker should be changed to `/perception/object_recognition/detection/clustering/objects`, or this pointcloud processing pipeline should be removed.

Anyway, the issue is out of scope of this PR. I confirmed other setting works after this PR.
LGTM

yukkysaito (approved):

YoshiRi (approved):

mojomex (approved):

technolojin:
@badai-nguyen 

Fix PR https://github.com/autowarefoundation/autoware_launch/pull/1960 WIP

PR #1960, opened 2026-08-25
fix(autoware_perception_launch): connect rule-based clusters to the tracker and split the lidar pipeline by clustering backend

technolojin:
## Description

The lidar pipeline launched the rule-based clustering chain on every run but never consumed its output: the tracker's `detection03` slot pointed at the PTv3 topic, which has no publisher in this pipeline, with its channel set to `none`.

- **Connect the clusters.** `detection03` follows the active clustering backend on channel `lidar_clustering`.
- **Restore the object filter.** `use_object_filter` (default `true`) and `detected_objects_filter_method` (default `lanelet_filter`) return to `component_perception.launch.xml`; the filter sits between the rule detector and the tracker. PTv3 clustering is label-based and stays unfiltered.
- **Split by clustering backend.** `object_recognition_lidar.launch.xml` hosts the rule-based pipeline; new `object_recognition_lidar_ptv3.launch.xml` hosts the PTv3 pipeline. `component_perception.launch.xml` selects the launcher instead of branching inside.
- **Launcher head.** `component_perception.launch.xml` follows the sibling launcher style: topic definitions in a head interfaces block, one `ground_segmentation_backend` let driving three sibling groups.
- **Fold the outlier filter.** The rule-based ground removal launcher runs the occupancy grid map outlier filter as its own composable node; the single-frame / time-series filter launchers and `use_obstacle_segmentation_*_filter` args are removed.

```
euclidean_cluster → shape_estimation → feature_remover
  → clustering/unfiltered/objects → object_lanelet_filter → clustering/objects
  → tracker detection03, channel lidar_clustering
```

## How was this PR tested?

For the clustering commits, a script expanded the full include tree and diffed every node's remappings and channel parameters against `main` over a 24-combination matrix: PTv3 combinations byte-identical, rule-based combinations differ only in the cluster path. The outlier filter fold was not re-run through that matrix. `pre-commit` passes. Not tested on vehicle or rosbag.

* rule based cluster 

Screenshot from 2026-08-26 14-28-01

* ml (ptv3)

Screenshot from 2026-08-26 13-50-45

## Notes for reviewers

`object_recognition_camera_lidar_radar.launch.xml` has the same dangling `detection03` subscription with `use_object_filter:=false`. Out of scope here.

## Effects on system behavior

Rule-based cluster objects now reach the tracker on `lidar_clustering`, lanelet-filtered by default. That channel has `can_spawn_new_tracker: true`, so expect extra tracks where the DNN misses objects. `use_object_filter:=false` bypasses the filter.

The rule-based ground segmentation always runs the occupancy grid map outlier filter; the single-frame filter option is gone. The PTv3 pipeline is unchanged.


autoware_universe_launch/autoware_perception_launch/launch/obstacle_segmentation/ground_segmentation_rule_based.launch.xml
```diff
@@ -31,11 +38,22 @@
...
+      pkg="autoware_occupancy_grid_map_outlier_filter"
+      plugin="autoware::occupancy_grid_map_outlier_filter::OccupancyGridMapOutlierFilterComponent"
+      name="occupancy_grid_based_outlier_filter"
+    >
+      <remap from="~/input/occupancy_grid_map" to="$(var input/occupancy_grid_map)"/>
+      <remap from="~/input/pointcloud" to="$(var output/single_frame_pointcloud)"/>
+      <remap from="~/output/pointcloud" to="$(var output/pointcloud)"/>
+      <param from="$(var occupancy_grid_map_outlier_filter_node_param_path)"/>
+      <extra_arg name="use_intra_process_comms" value="$(var use_intra_process)"/>
+    </composable_node>
```
YoshiRi: [Q]
Do we really need ogm outlier filter now? I think we can simply skip this if we use this in real environment for computational reason

technolojin: removed 
https://github.com/autowarefoundation/autoware_launch/pull/1960/changes/f5690eb725628829f9bbef123f408477cb1ac57c

YoshiRi (approved):
LGTM

PR #1962, opened 2026-08-27
refactor(control)!: migrate tier4_control_launch to autoware_control_launch with a separate config package

technolojin:
## Description

Part of #1939 (control item), supporting
[discussion #7111](https://github.com/orgs/autowarefoundation/discussions/7111) and #1848.

Moves the control launch stack out of the product layer and splits its parameter tree into a
dedicated package, so that a platform can supply its own control parameters without forking the
launchers. Sibling of #1954 (planning), following the same package boundary. Built on #1963 (code owner housekeeping, merged).

This is the migration step only: the control launch graph itself (`control_container` with the
command gate, mode transition manager and trajectory follower; `control_check_container` with the
optional checkers; the evaluator) is unchanged. Decomposition into reusable feature units follows in
#1965, stacked on this branch.

### What was done

1. **Package rename**: `tier4_control_launch` → `autoware_universe_launch/autoware_control_launch`,
   and the old package is removed.
2. **Separate config package**: the control parameter tree moves from
   `autoware_launch/config/control/` into a new data-only package
   `autoware_universe_launch/autoware_control_config`, dropping the redundant `control/` level.
3. **Config addressed by argument**: `control.launch.xml` resolves the whole tree through
   `control_config_pkg` (default `autoware_control_config`). Platforms that differ in only a few
   files instead override the individual `*_param_path` arguments, which are now all declared at
   the top of `control.launch.xml` with defaults.
4. **Module preset moved in**: `control.launch.xml` includes the preset itself and takes
   `module_preset`. Values passed in by a caller still win, since the include is the first statement.
   The config root is resolved once (`control_config_dir`) and shared by the preset include and the
   parameter paths.
5. **Per-vehicle controller params stay product-side**: `control.launch.xml` takes no `vehicle_id`.
   `use_individual_control_param` / `vehicle_id` live in the component, which resolves
   `trajectory_follower/<vehicle_id>/` and passes `lat_controller_param_path` /
   `lon_controller_param_path` down; the launcher only sees file paths.
6. **Thin system component**: `components/tier4_control_component.launch.xml` (69 lines) becomes
   `components/component_control.launch.xml` (49 lines) carrying only the product-level junctions,
   boundary topics, the per-vehicle path selection and the global parameter loader. It exposes
   `control_config_pkg` (same shape as `perception_config_pkg`) and forwards it, so a platform
   relocates the whole tree from one place. `vehicle_id` and `check_external_emergency_heartbeat`
[...]

PR #1971, opened 2026-08-31
refactor(localization)!: migrate tier4_localization_launch to autoware_localization_launch with a separate config package

technolojin:
## Description

Part of #1939 (localization item), supporting
Sibling of #1954 (planning) and #1962 (control), following the same package boundary.

Moves the localization launch stack out of the product layer and splits its parameter tree into a
dedicated package, so a platform can supply its own localization parameters without forking the
launchers. The launch graph — node names, namespaces, topics — is unchanged.

### Requirements

- autowarefoundation/autoware_core#1409 (merged) — the `PoseInitializer.node` interfaces used in item 5.
- autowarefoundation/autoware_universe#13298 — soft dependency: `AutomaticPoseInitializer.node.yaml` declares
  `initialization_state` and `initialize` as `remap_target` instead of `global`. Until it lands, the
  designer skips the two `LocalizationUtil` connections added in item 5 (a global input port cannot be
  a connection target) and reports them as build warnings. The exported launch graph is the same either
  way, so this PR does not depend on it to run. (Previously split out as #13299, which is now folded
  back into #13298.)

### What was done

1. **Packages**: `tier4_localization_launch` → `autoware_universe_launch/autoware_localization_launch`;
   `autoware_launch/config/localization/` → the new data-only
   `autoware_universe_launch/autoware_localization_config` (the redundant `localization/` level dropped).
2. **Config addressed by argument**: every unit launcher declares its own `*_param_path` defaults and
   resolves them through `localization_config_pkg` (default `autoware_localization_config`). The
   component declares the argument once; the units inherit it, as in `component_perception`. A system
   only ever references the parameters of the units it includes.
3. **Mode selection flattened into the product layer**: `pose_source`/`twist_source` string parsing,
   `stop_check_enabled` (from `system_run_mode`) and `user_defined_initial_pose/*` (from `initial_pose`)
   were hidden derivations inside `pose_twist_estimator.launch.xml`. They move into
   `components/component_localization.launch.xml`, which exposes one flat junction per estimator
   (`use_ndt_pose`, `use_yabloc_pose`, …) so a system-designer-fixed system can pin the executing nodes
   directly, as perception and control do. `stop_check_enabled` now covers every run mode — the stop
   check belongs to a real vehicle, so anything but `online` disables it, where previously a run mode
   outside `online`/`logging_simulation` left the value undefined.
4. **Composition on the system side, perception-style**: the all-mode entry launchers
   (`localization.launch.xml`, `pose_twist_estimator.launch.xml`) are removed. The launch package ships
   one self-contained launcher per unit, and `component_localization.launch.xml` is the composition
   root — it includes only the units the selected sources need, pushes the namespaces and wires the
   boundary topics, mirroring how `Localization.module.yaml` composes its instances. The arbiter and
   the pose covariance modifier are launched from there, so `autoware_launch` takes their `exec_depend`.
   The unreferenced `Tier4LocalizationWrapper.node` is dropped, as perception's wrapper was.
5. **Design modules move alongside** the launch package, matching planning and control, and
   `LocalizationUtil` gains the `automatic_pose_initializer` ↔ `pose_initializer` connections that
   `util.launch.xml` has always launched.
6. **Consumers retargeted**: `tier4_simulator_component.launch.xml`, `tier4_simulator_launch`,
   `sync-params.yaml`, `CODEOWNERS-manual` and the guideline check. Both package structure diagrams
   are dropped: they draw the `localization.launch.xml` composition removed here, and the
   `autoware_launch` one also draws the simulator wrappers as if the root launcher called the sub
   launchers.

### Breaking changes

- `tier4_localization_launch` → `autoware_localization_launch`
- `autoware_launch/config/localization/**` → `autoware_localization_config/config/**` (no `localization/` level)
- `components/tier4_localization_component.launch.xml` → `components/component_localization.launch.xml`
- the component argument `loc_config_path` (a directory) is replaced by `localization_config_pkg`
  (a package name); individual files are overridden via the `*_param_path` arguments
- `localization.launch.xml` and `pose_twist_estimator.launch.xml` are removed: callers either use
  `component_localization.launch.xml` (which keeps `pose_source`/`twist_source`) or compose the unit
  launchers with the per-estimator `use_*` junctions
- `lidar_marker_localizer.launch.xml` renames `lidar_marker_localizer/<name>_param_path` → `<name>_param_path`
- `Tier4LocalizationWrapper.node` is removed (no system referenced it; systems compose `Localization.module`)

### Verification

- `colcon build` of `autoware_localization_launch`, `autoware_localization_config`, `autoware_launch`,
  `tier4_simulator_launch` and `autoware_sample_designs` (designer deploy) pass; `pre-commit` clean.
- Launch-graph equivalence against `main`: resolved node dumps (package, executable/plugin, final node
  name, parameters with generated files inlined, remaps, target containers) are **identical** for
  `pose_source:=ndt` (12 records) and `ndt_yabloc` (26 records), modulo the parameter-file relocation —
  re-verified after the junction flattening and after the unit-launcher recomposition.
- Logging simulator and E2E simulator (simulators using localization component) are tested. 

### Kept at this stage

Decomposition of the localization units into finer reusable feature units is a follow-up, as control
did in #1965.

`tier4_simulator_component.launch.xml` still addresses the six localization parameter files through a
hardcoded `$(find-pkg-share autoware_localization_config)` instead of `localization_config_pkg`, so a
product that relocates the tree still gets the stock parameters for the planning simulator's
`ekf_localizer`, `stop_filter`, `twist2accel`, `pose_instability_detector`, `pose_initializer` and the
error monitor. Routing them through the argument belongs with the simulator component's own migration,
not here.


.github/sync-params.yaml
```diff
@@ -119,7 +119,7 @@ perception:
           - path: autoware_universe_launch/autoware_perception_config/config/object_recognition/tracking/tracking_object_merger/decorative_tracker_merger.param.yaml
       # Perception/obstacle_segmentation
       - source: perception/autoware_elevation_map_loader/config/elevation_map_parameters.yaml
-        # note: this is not a valid ROS2 param file format.
+        # note: this is not a valid ROS 2 param file format.
```
technolojin, line 122: it is to avoid existing CI warning

technolojin:
> https://github.com/autowarefoundation/autoware_launch/pull/1972 replaces this PR

wrong PR closed. reopen this PR.

takayuki5168 (approved):

shmpwk (approved):

YamatoAndo (approved):

shmpwk:
@mitsudome-r @mojomex Could you review this PR :bow:

maxime-clem (approved):

mitsudome-r (approved):

PR #1972, opened 2026-09-01
fix(control, planning): connect the module inputs left without a source

technolojin:
## Description

Connects the module inputs the design graph left without a source, in `Planning.module` and `Control.module`. Each reproduces wiring the launch files already perform; the design had not encoded it, so the exported launcher dropped the remap and the node subscribed to an unresolved relative name.

**Planning**

- `remaining_distance_time_calculator.planning_velocity` ← `scenario_planning.current_max_velocity`
- `external_velocity_limit_selector` internal limit and clear command ← `motion_velocity_planner`, forwarded through `MotionPlanning` and `LaneDriving`. The selector previously had no input remaps at all, so the external velocity limit path was inert.
- `manual_lane_change_handler`: route availability, and the `set_preferred_primitive` client
- `lane_driving.costmap` ← `parking` (new `occupancy_grid` port); the `costmap` input on `ScenarioPlanning` is dropped, since the producer is inside the module
- `MissionPlanning.route_marker` ← `mission_planner`, restoring the topic `autoware.rviz` expects

**Control**

- `collision_detector.operation_mode_state` ← `api_operation_mode_state`
- operation mode transition manager: gate mode, selector mode, and the `select_external_command` client
- `stop_mode_operator`: steering and velocity

New component interfaces, where the counterpart is outside Planning/Control: `Control.client.control_mode_request` (vehicle interface), `Planning.server.set_preferred_lane` (AD API), `Planning.subscriber.system_hazard_lights_cmd` (MRM operator).

## Related links

- autowarefoundation/autoware_core#1416 — velocity limit message types. **Required**: without it the `current_max_velocity -> planning_velocity` connection fails the designer's type check.
- autowarefoundation/autoware_universe#13298 — the matching node design updates. **Required** by the `system_hazard_lights_cmd` connection. (Previously split out as #13304, which is now folded back into #13298.)

## How was this PR tested?

- `autoware_sample_designs` builds clean with both linked PRs in the workspace; designer warning set unchanged.
- Diffed the exported `Runtime/main_ecu` launcher: five previously dropped remaps now appear, with the legacy absolute names preserved. Three topics are renamed by design, including `route_marker` → `/planning/mission_planning/route_marker`, which fixes an RViz mismatch.
- `pre-commit` on every changed design file.

## Notes for reviewers

Design-only change; no launch file or source code is touched.

Conflicts with #1965, which splits `Control.module.yaml` into sub-modules — whichever lands first, the other rebases.

Left unconnected on purpose: `behavior_velocity_planner.external_velocity_limit_mps` (API topic with no in-design publisher; belongs node-side in `autoware_core`), `planning_evaluator.reference_trajectory` (a launch-time `<let>` switch, not fixed topology), and `stop_mode_operator`'s outputs (their consumer `control_command_gate` is not modeled as a design node).

## Interface changes

New ports: `Control.client.control_mode_request`, `Planning.server.set_preferred_lane`, `Planning.subscriber.system_hazard_lights_cmd`, `Parking.publisher.occupancy_grid`, `clear_velocity_limit` on `MotionPlanning` and `LaneDriving`. Removed: `ScenarioPlanning.subscriber.costmap`.

## Effects on system behavior

Restores the external velocity limit path, the operation mode transition manager's gate and selector inputs, the manual lane change handler's route availability and preferred-primitive service, and the behavior planner's costmap input — all of which resolved to unbound relative names in the exported launcher.

### DIFF
--- a/autoware_sample_designs/design/module/sensing/SampleSensorKit.module.yaml
+++ b/autoware_sample_designs/design/module/sensing/SampleSensorKit.module.yaml
@@ -17,6 +17,7 @@ subscribers:
   - name: odometry
   - name: twist
   - name: pose_ndt
+  - name: map_projector_info
 
 publishers:
   - name: imu/imu_data
@@ -39,6 +40,8 @@ connections:
   - - imu.publisher.imu_data
     - publisher.imu/imu_data
   # GNSS
+  - - subscriber.map_projector_info
+    - gnss.subscriber.map_projector_info
   - - gnss.publisher.pose
     - publisher.gnss/pose
   - - gnss.publisher.pose_with_covariance
--- a/autoware_sample_designs/design/module/sensing/SampleSensorKitGnss.module.yaml
+++ b/autoware_sample_designs/design/module/sensing/SampleSensorKitGnss.module.yaml
@@ -10,6 +10,7 @@ instances:
 
 subscribers:
   - name: orientation
+  - name: map_projector_info
 publishers:
   - name: pose
   - name: pose_with_covariance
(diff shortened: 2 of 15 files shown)


technolojin:
replaced by https://github.com/autowarefoundation/autoware_launch/pull/1976

PR #1980, opened 2026-09-10
refactor(map)!: migrate tier4_map_launch to autoware_map_launch with a separate config package

shmpwk:
## Description

Part of #1939 (map item), supporting
Follows #1971 (localization) and its siblings #1954 (planning) / #1962 (control), using the same
package boundary. Builds on #1970, which composed the map component from `Map.module`.

Moves the map launch stack out of the product layer and splits its parameter tree into a dedicated
package, so a platform can supply its own map parameters without forking the launchers. The launch
graph — node names, namespaces, topics, parameters — is unchanged.

### What was done

1. **Packages**: `tier4_map_launch` → `autoware_universe_launch/autoware_map_launch`;
   `autoware_launch/config/map/` → the new data-only `autoware_universe_launch/autoware_map_config`
   (the redundant `map/` level dropped).
2. **Config addressed by argument**: every unit launcher declares its own `*_param_path` defaults and
   resolves them through `map_config_pkg` (default `autoware_map_config`). The component declares the
   argument once; the units inherit it, as in `component_perception` and `component_localization`.
   A system only ever references the parameters of the units it includes.
3. **One self-contained launcher per unit**: the all-mode entry launcher `map.launch.xml` is removed
   and the launch package ships one launcher per map feature — `map_loader/pointcloud_map_loader`,
   `map_loader/lanelet2_map_loader` (loader plus its marker visualization),
   `map_projection_loader/map_projection_loader`, `map_tf_generator/vector_map_tf_generator` and
   `map_hash_generator/map_hash_generator`. Each takes `ld_preload_value` as an argument, as the
   control units do, instead of reaching for a variable the entry launcher happened to leave in scope.
4. **Composition in the product layer**: `autoware_launch/launch/components/component_map.launch.xml`
   is the composition root — it derives the five map file paths from `map_path` /
   `lanelet2_map_file` / `pointcloud_map_file`, resolves the executor once, pushes the `map`
   namespace and includes the units, mirroring how `Map.module` composes its instances. It exposes one
   flat junction per unit (`use_pointcloud_map`, `use_lanelet2_map`, `use_map_projection`,
   `use_map_tf_generator`, `use_map_hash_generator`, all `true`) so a system-designer-fixed system can
   pin the executing nodes directly, as perception, control and localization do.
5. **Design module moves alongside** the launch package (`design/module/Map.module.yaml`), matching
   planning, control and localization. The unreferenced `Tier4MapWrapper.node` is dropped, as
   perception's and localization's wrappers were — no system referenced it since #1970.
6. **Consumers retargeted**: `autoware.launch.xml`, `autoware_launch/package.xml`,
   `universe_map.parameter_set.yaml`, `sync-params.yaml` and `CODEOWNERS-manual`. Both package
   structure diagrams are dropped: `map_launch.drawio.svg` draws the `map.launch.xml` composition
   removed here (as `map.launch.py`, which has not existed for some time), and the `autoware_launch`
   one draws the same box plus the simulator wrappers as if the root launcher called the sub
   launchers. #1971 removes the latter for the same reason.
7. **Missing dependency added**: `autoware_lanelet2_map_visualizer` is now an `exec_depend` of the
   launch package. `tier4_map_launch` launched the node without declaring it.

### Breaking changes

- `tier4_map_launch` → `autoware_map_launch`
- `autoware_launch/config/map/**` → `autoware_map_config/config/**` (no `map/` level)
- `components/tier4_map_component.launch.xml` → `components/component_map.launch.xml`
- `map.launch.xml` is removed: callers either use `component_map.launch.xml` (which keeps `map_path`,
  `lanelet2_map_file`, `pointcloud_map_file`) or compose the unit launchers themselves. Callers that
  passed the five resolved map paths (`pointcloud_map_path`, `pointcloud_map_metadata_path`,
  `lanelet2_map_path`, `lanelet2_map_metadata_path`, `map_projector_info_path`) now either let the
  component derive them from `map_path` or pass them to the individual unit launchers.
- the four `*_param_path` arguments keep their names but default to `autoware_map_config`; the whole
  tree is relocated through `map_config_pkg`
- `Tier4MapWrapper.node` is removed (no system referenced it; systems compose `Map.module`)

### Verification

- `pre-commit` clean over the changed files, including `check package depends`, `prettier launch.xml`
  and the system-design-format lint.
- Launch-graph equivalence against `main`: resolved node dumps (package, executable, final node name,
  remaps, `LD_PRELOAD`, and parameter files with the generated substitution files inlined) are
  **identical** between `tier4_map_component.launch.xml` on `main` and `component_map.launch.xml` on
  this branch — 6 records, byte-for-byte. Each junction was checked to drop exactly its own unit, and
  `map_config_pkg` was checked to relocate the tree from the component down into the units.

### Kept at this stage

`lanelet2_map_visualization` ships inside `lanelet2_map_loader.launch.xml` rather than as its own
unit: it exists only to render that loader's output, and splitting it would add a junction with no
system behind it yet. Decomposition into finer reusable feature units is a follow-up, as control did
in #1965.


YamatoAndo (approved):

shmpwk:
@technolojin Could you review this :pray:

KYabuuchi (approved):

mitsudome-r (approved):

technolojin (approved):

PR #1988, opened 2026-09-15
feat(autoware_localization_launch): run random_downsample_filter standalone when agnocast is enabled

Koichi98:
## Description

Companion launch change for autowarefoundation/autoware_universe#13386, which moves `RandomDownsampleFilterComponent` to `agnocast_wrapper::Node`. Since `agnocast_wrapper::Node` is not designed to run inside a component container intentionally, `random_downsample_filter` is started as a standalone process when `ENABLE_AGNOCAST=1` and loaded into `localization_pointcloud_container_name` otherwise. The node name, namespace, remaps and parameters are the same in both branches.

## How was this PR tested?

Ran `logging_simulator.launch.xml` on `sample-map-rosbag` with the `sample-rosbag` bag, with `ENABLE_AGNOCAST=0` and `=1`. In both modes the whole downsampling chain keeps running.
## Notes for reviewers

`<composable_node>` accepts `if` / `unless`, so the branch needs one condition attribute rather than a second `<load_composable_node>` block. `crop_box_filter_measurement_range` and `voxel_grid_downsample_filter` in the same block are on the same migration path (autoware_universe#13384 and #13072), so this block is expected to lose its remaining nodes later.

## Effects on system behavior

None with `ENABLE_AGNOCAST` unset or `0`. With `ENABLE_AGNOCAST=1`, `random_downsample_filter` runs as a standalone process instead of a composable node; topic names, types and QoS are unchanged.


YoshiRi (approved):
LGTM

